PR #50, opened 2021-07-07
Update deps

osmGosha:
Version 0.7.0


osmGosha:
> Can we use https://www.npmjs.com/package/json-sql?

We use it, but it is now located in the /modules folder with updated dependencies. The npm package is not updated and contains vulnerabilities.

dev-adamant-im:
Bullshit.

PR #61, opened 2022-11-23
fix: permissions for adamant_main db in postgre 15

adamant-al:
```
adamant@server1:~/adamant$ node app.js
[FTL] 2022-11-22 15:54:58 | Error: permission denied for schema public
```

https://stackoverflow.com/questions/74110708/postgres-15-permission-denied-for-schema-public

Fix:
```
ALTER DATABASE adamant_main OWNER TO adamant;
```


martiliones (approved):
makes sense

PR #62, opened 2023-01-14
Fix CORS error for socket.io and add CORS config

martiliones:
`cors` options for `config.json`:

* `origin`: Configures the **Access-Control-Allow-Origin** CORS header. Possible values:
  - `Boolean` - set `origin` to `true` to reflect the [request origin](http://tools.ietf.org/html/draft-abarth-origin-09), as defined by `req.header('Origin')`, or set it to `false` to disable CORS.
  - `String` - set `origin` to a specific origin. For example if you set it to `"http://example.com"` only requests from "http://example.com" will be allowed.
  - `Array` - set `origin` to an array of valid origins. For example `["http://example1.com", "http://example2.com/"]` will accept any request from "http://example1.com" or "example2.com".
* `methods`: Configures the **Access-Control-Allow-Methods** CORS header. Expects a comma-delimited string (ex: 'GET,PUT,POST') or an array (ex: `['GET', 'PUT', 'POST']`).

default is `{ origin: true }` - allows any origins

### DIFF
--- a/app.js
+++ b/app.js
@@ -218,6 +218,10 @@ d.run(function () {
         throw Error(e);
       }
 
+      if (appConfig.cors) {
+        appConfig.cors.credentials = true;
+      }
+
       if (appConfig.dapp.masterrequired && !appConfig.dapp.masterpassword) {
         var randomstring = require('randomstring');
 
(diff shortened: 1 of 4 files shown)


adamant-al (approved):

adamant-al:
Update Wiki.

PR #66, opened 2023-01-24
[feat] add logs when executing migrations

martiliones:


helpers/database.js
```diff
@@ -108,6 +108,10 @@ function Migrator (pgp, db) {
    * @return {function} waterCb with error | appliedMigrations
    */
   this.applyPendingMigrations = function (pendingMigrations, waterCb) {
+    if (pendingMigrations.length) {
+      logger.info(`Found ${pendingMigrations.length} pending migrations. Start executing, this may take a while.`);
```
adamant-al, line 112:
```suggestion
      logger.info(`Found ${pendingMigrations.length} pending migrations. Start executing, this may take a while…`);
```

helpers/database.js
```diff
@@ -120,6 +124,10 @@ function Migrator (pgp, db) {
         return eachCb(err);
       });
     }, function (err) {
+      if (pendingMigrations.length && !err) {
+        logger.info('Migrations have been successfully completed');
```
adamant-al, line 128:
```suggestion
        logger.info('Migrations have been successfully completed.');
```

adamant-al: If case of error, will it be a message?

martiliones: Yes, you will see a fatal message in this case

adamant-al (approved):

PR #68, opened 2023-04-11
feat: add default config

martiliones:
By adding `config.json` to `.gitignore`, we ensure that sensitive configuration data is not accidentally committed to the repository. Additionally, it makes easier to pull changes and update the Node on a server.


adamant-al:
- Add changes to `install_node_centos.sh` as well
- Check if file `config.default.json` exists for backward compatibility. If no such a file, use `config.json`. See https://medium.com/adamant-im/how-to-run-your-adamant-node-on-ubuntu-990e391e8fcc#4934

martiliones:
isn't the installation script meant to download the __latest__ version of adamant? I don't understand how the `default.config.json` could be missing.

adamant-al:
If we decide to merge `dev` into `master` right now, that will work.

adamant-al (approved):

PR #79, opened 2024-06-20
Feat: find a delegate by the given address

martiliones:
[`/delegates/get?address=U12693243886978615076`](https://ahead.adamant.im/api/delegates/get?address=U12693243886978615076)

additionally, optimized `/delegates/get` endpoint to search for a single transaction

Trello: Optimize /delegates/get and add address parameter https://trello.com/c/1OTebrME

### DIFF
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -417,6 +417,7 @@ Delegates.prototype.generateDelegateList = function (height, cb) {
  * Gets delegates and for each one calculates rate, rank, approval, productivity.
  * Orders delegates as per criteria.
  * @param {Object} query
+ * @param {Object} filter - account specific filters, e.g. `username`, `publicKey` or `address`
  * @param {function} cb - Callback function.
  * @return {setImmediateCallback} error| object with delegates ordered, offset, count, limit.
  * @todo OrderBy does not affects data? What is the impact?.
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -765,13 +765,19 @@ Delegates.prototype.shared = {
       const filter = {}
 
       if (req.body.publicKey) {
-        filter.publicKey = req.body.publicKey;
+        filter.address = modules.accounts.generateAddressByPublicKey(
+          req.body.publicKey,
+        );
       } else if (req.body.username) {
         filter.username = req.body.username;
       } else if (req.body.address) {
         filter.address = req.body.address;
       } else {
         return setImmediate(cb, 'Delegate not found');
+      }
+
+      if (req.body.address && req.body.publicKey && filter.address !== req.body.address) {
+        return setImmediate(cb, 'Delegate publicKey does not match address');
       }
 
       modules.delegates.getDelegates(req.body, filter, function (err, data) {
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -421,15 +421,22 @@ Delegates.prototype.generateDelegateList = function (height, cb) {
  * @return {setImmediateCallback} error| object with delegates ordered, offset, count, limit.
  * @todo OrderBy does not affects data? What is the impact?.
  */
-Delegates.prototype.getDelegates = function (query, cb) {
+Delegates.prototype.getDelegates = function (query, filter, cb) {
   if (!query) {
     throw 'Missing query argument';
   }
+
+  if (typeof filter === 'function') {
+    cb = filter
+    filter = {}
+  }
+
   var sortFilter = { 'vote': -1, 'publicKey': 1 };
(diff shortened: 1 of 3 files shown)


adamant-al (approved):

modules/delegates.js
```diff
@@ -421,15 +421,22 @@ Delegates.prototype.generateDelegateList = function (height, cb) {
...
-Delegates.prototype.getDelegates = function (query, cb) {
+Delegates.prototype.getDelegates = function (query, filter, cb) {
   if (!query) {
     throw 'Missing query argument';
   }
+
+  if (typeof filter === 'function') {
+    cb = filter
+    filter = {}
+  }
```
bludnic, line 432: Is this really necessary? Dynamic function arguments can be confusing.

Just pass undefined if you don't need to apply a filter. I think this approach is more clear.
```js
modules.delegates.getDelegates(query, undefined, cb)
```

bludnic: A good solution would be to convert the function arguments into an object. This way, you can easily omit the `filter` property if it's not needed. However, this will create significant changes in the codebase so I would stick with the previous approach.

bludnic (changes requested):

modules/delegates.js
yoxira: It was made to make sure it doesn't break anything, but it seems like the method is not commonly used.

modules/delegates.js
```diff
@@ -421,15 +421,17 @@ Delegates.prototype.generateDelegateList = function (height, cb) {
  * @return {setImmediateCallback} error| object with delegates ordered, offset, count, limit.
  * @todo OrderBy does not affects data? What is the impact?.
  */
-Delegates.prototype.getDelegates = function (query, cb) {
+Delegates.prototype.getDelegates = function (query, filter, cb) {
```
bludnic, line 424: Can you describe the `filter` param in JSDoc?

modules/delegates.js
```diff
@@ -759,20 +762,24 @@ Delegates.prototype.shared = {
         return setImmediate(cb, err[0].message);
       }
 
-      modules.delegates.getDelegates(req.body, function (err, data) {
+      const filter = {}
+
+      if (req.body.publicKey) {
+        filter.publicKey = req.body.publicKey;
```
bludnic, line 768: It seems like the find delegate by public key is broken:

http://localhost:36666/api/delegates/get?publicKey=e8ecea9faf2b9cfa26d635bc94d8457114beafc90b4f6b5a7f1354d193ab0f3b

image

Although same request on production node works fine:

https://bid.adamant.im/api/delegates/get?publicKey=e8ecea9faf2b9cfa26d635bc94d8457114beafc90b4f6b5a7f1354d193ab0f3b

image

modules/delegates.js
```diff
@@ -759,20 +762,24 @@ Delegates.prototype.shared = {
...
+      } else if (req.body.username) {
+        filter.username = req.body.username;
+      } else if (req.body.address) {
+        filter.address = req.body.address;
```
bludnic, line 772: Shouldn't `req.body` params be sanitized? Is SQL injection possible here? @martiliones

bludnic (changes requested):

modules/delegates.js
```diff
@@ -759,20 +762,30 @@ Delegates.prototype.shared = {
         return setImmediate(cb, err[0].message);
       }
 
-      modules.delegates.getDelegates(req.body, function (err, data) {
+      const filter = {}
+
+      if (req.body.publicKey) {
```
bludnic, line 767: GET http://localhost:36666/api/delegates/get?publicKey=qwe

`500 Internal Server Error`
```json
{"success":false,"error":"API error: The argument 'encoding' is invalid for data of length 3. Received 'hex'"}
```

GET https://bid.adamant.im/api/delegates/get?publicKey=qwe

`200 OK`

```json
{"success":false,"nodeTimestamp":224017224,"error":"Delegate not found"}
```

Expected same behaviour. @martiliones

bludnic:
Don't forget to update the Wiki after merge
https://github.com/Adamant-im/adamant/wiki/API-Specification#get-delegate

adamant-al:
@yoxira 
Update

modules/delegates.js
yoxira: Added validation for `publicKey` and `address` properties. It's not the same as `Delegate not found` but the right way to handle this, in my opinion.

bludnic:
Please update also schema.adamant.im @yoxira

yoxira:
@bludnic 
https://github.com/Adamant-im/adamant-schema/pull/15

bludnic (approved):

PR #80, opened 2024-06-20
Feat/ws transaction types

martiliones:
### Fixed
- A memory leak when subscribing to multiple addresses
- A bug when the websocket would send a transaction that has address which starts with the subscribed address 

### Added

Usage:

```js
// w/o an address
socket.emit('types', [TransactionType.SEND, TransactionType.CHAT_MESSAGE]);
socket.on('newTrans', console.log.bind('Chat transaction without bounding to a specific address:'));
```
```js
// w/ an address
socket.emit('address', 'U1234567890123456');
socket.emit('types', [TransactionType.SEND, TransactionType.CHAT_MESSAGE]);
socket.on('newTrans', console.log.bind('Chat transaction for the user:'));

```

### DIFF
--- a/modules/clientWs.js
+++ b/modules/clientWs.js
@@ -16,7 +16,7 @@ class ClientWs {
     this.logger = logger;
     io.sockets.on('connection', (socket) => {
       try {
-        const describe = new TransactionSubscription();
+        const describe = new TransactionSubscription(socket);
 
         socket.on('address', (address) => {
           const addresses = Array.isArray(address) ? address : [address];
@@ -57,7 +57,7 @@ class ClientWs {
     try {
       const subs = findSubs(t, Object.values(this.describes));
       subs.forEach((s) => {
-        s.emit('newTrans', t);
+        s.socket.emit('newTrans', t);
       });
     } catch (e) {
       this.logger.debug('Socket error emit ' + e);
--- a/modules/clientWs/transactionSubscription.js
+++ b/modules/clientWs/transactionSubscription.js
@@ -7,7 +7,9 @@ const {
 const validator = new ZSchema({noEmptyStrings: true});
 
 class TransactionSubscription {
-  constructor() {
+  constructor(socket) {
+    this.socket = socket;
+
     /**
      * List of addresses to subscribe to
      * @type {Set<string>}


adamant-al:
@martiliones 
Create a Trello card for the task

adamant-al:
> should we allow unsubscribing from types?

Provide a reason and an example.

bludnic:
> you can test using the following code snippet and `ahead` node:
> 
> ```js
> import {io} from 'socket.io-client';
> 
> const connection = io('https://ahead.adamant.im/', {
>   reconnection: false,
>   timeout: 5000
> });
> 
> connection.on('connect', () => {
>   console.log('connected');
> 
>   connection.emit('types', [0, 8]);
> });
> 
> connection.on('newTrans', (transaction) => {
>   console.log(transaction)
> });
> 
> connection.on('disconnect', (reason) => {
>   console.log(`disconnected, reason: ${reason}`)
> })
> 
> connection.on('connect_error', error =>
>   console.warn(`connection error: ${error}`)
> );
> ```

It doesn't work for me. After successful connection it does not emit new transactions. Tried to send a message and in-chat transfer.

image

bludnic:
As I see previously we could subscribe to several addresses by emitting "address" multiple times, but now since you changed the ID of the subscription this is not possible.

Previously:

```js
import { io } from 'socket.io-client';

const connection = io('https://debate.adamant.im/', {
  reconnection: false,
  timeout: 5000
});

connection.on('connect', () => {
  console.log('connected');

  connection.emit('address', 'U9203183357885757380')
  connection.emit('address', 'U14724553101605161229')
  connection.emit('types', [0, 8]);
});

connection.on('newTrans', (transaction) => {
  console.log(transaction)
});
```

I think using socket.id as an ID is a good approach, but we must keep the previous functionality. Therefore, I suggest passing "address" as an array. To maintain backward compatibility, I would also handle the string value type.

```js
connection.on('connect', () => {
  console.log('connected');

  connection.emit('address', ['U9203183357885757380', 'U14724553101605161229'])
  connection.emit('types', [0, 8]);
});
```

modules/clientWs.js
```diff
@@ -15,15 +15,36 @@ class ClientWs {
...
+           * Flags for subscribing to transaction types.
+           * Each bit represents a transaction type. For example:
+           *
+           * 0b100000001
+           *   │       └ transfer transaction (type - 0)
+           *   │
+           *   └─ message transaction (type - 8)
+           */
+          types: 0,
+        };
```
bludnic, line 30: Can we use an array of numbers for simplicity? What are the benefits of your approach?

yoxira: much better memory efficiency and speed, obviously

yoxira:
> It doesn't work for me

I have reverted the ahead node commits in favor of testing #82

modules/clientWs.js
bludnic: I agree that is more efficient, but does we really win a lot in this specific case?

adamant-al: @yoxira 
Note than soon we'll have more tx types.

adamant-al:
@yoxira 
See the Testing blockchain dev updates https://docs.google.com/document/d/15LWxiLf6VQin3cT8UFEoJd9mng-sD4JJpr5KP98J82c/edit#heading=h.bmjmb6vgdt7k
Waiting for raising height 0 ⟶ current on the ahead node may take a month. Not acceptable.

modules/clientWs.js
```diff
@@ -65,11 +86,18 @@ function getUTime () {
...
   const filterred = [];
   for (let aId in subs) {
-    if (aId.startsWith(address1) || aId.startsWith(address2)) {
-      filterred.push(subs[aId]);
+    const sub = subs[aId];
+    const {address, types, socket} = sub;
+    const isTypeAllowed = !types || types & (1 << type)
+
+    if (
+      ([address1, address2].includes(address) && isTypeAllowed) ||
```
yoxira, line 97: `[address1, address2].includes(address)` doesn't do anything, if it's false then it's just going to pass through because of `|| isTypeAllowed`

modules/clientWs.js
```diff
@@ -15,15 +15,36 @@ class ClientWs {
...
+        socket.on('address', (address) => {
+          if (typeof a === 'string') {
+            describe.address = address;
+            this.describes[socket.id] = describe;
+          }
         });
+
+        socket.on('types', (transactionTypes) => {
+          if (Array.isArray(transactionTypes)) {
+            transactionTypes.forEach((type) => (describe.types |= 1 << type));
```
yoxira, line 41: this may lead to unsafe integer

modules/clientWs.js
```diff
@@ -15,15 +15,36 @@ class ClientWs {
...
+           * 0b100000001
+           *   │       └ transfer transaction (type - 0)
+           *   │
+           *   └─ message transaction (type - 8)
+           */
+          types: 0,
+        };
+
+        socket.on('address', (address) => {
+          if (typeof a === 'string') {
```
yoxira, line 33: use RegExp to test the addresses

modules/clientWs.js
```diff
@@ -15,15 +15,36 @@ class ClientWs {
     this.logger = logger;
     io.sockets.on('connection', (socket) => {
       try {
-        let address = '';
-        let aId = '';
-        socket.on('address', (a) => {
-          address = a;
-          aId = address + '_' + socket.id;
-          this.describes[aId] = socket;
+        const describe = {
```
yoxira, line 18: use class instead?

yoxira (changes requested):

modules/clientWs.js
bludnic: Will your solution work if the transaction type cannot be stored in one byte, e.g. `type: 256`?

yoxira: It will work using up to 52 bits (=each bit representing transaction type)

yoxira: It works for types from `0` to `52`, if we'll ever need a higher type, we can just use bigint

yoxira: to save memory I assume we need to filter out invalid types / addresses anyway

bludnic: So, if we exceed the value of `52`, we should modify the code to keep it working, right? In that case, using an array of numbers is a universal solution.

helpers/tranasctionTypesBoundary.js
```diff
@@ -0,0 +1,8 @@
+const TransactionTypes = require('./transactionTypes');
+
+const transactionValues = Object
+  .values(TransactionTypes)
+  .filter((type) => typeof type === 'number');
```
yoxira, line 5: does this look unreliable?

yoxira:
- Added back subscribing to multiple addresses
- Rewritten to use `Set`

modules/clientWs/transactionSubscription.js
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * @returns {boolean} - whether succuessfuly subscribed to at least one address
+   */
+  subscribeToAddresses(...addresses) {
+    let subscribed = false;
+
+    addresses.forEach((address) => {
+      if (
+        typeof address === 'string' && address.length !== 0 &&
+        validator.validate(address, { format: 'address' })
```
bludnic, line 49: I see that the validator already has a check for [str.length](https://github.com/Adamant-im/adamant/blob/b2f4508c2f23816e423df7ec7c372fcff25bad2d/helpers/z_schema.js#L36).

It may be enough call the validator without additional checks:

```js
if (validator.validate(address, { format: 'address' })) { // ...
```

modules/clientWs/transactionSubscription.js
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @type {Set<string>}
+     */
+    this.addresses = new Set();
+
+    /**
+     * @type {Set<number>}
+     */
+    this.types = new Set();
+  }
```
bludnic, line 20: Add a comment referring to [transactionTypes.js](https://github.com/Adamant-im/adamant/blob/fd4d79f6588e1b729859ea2bc2b8ae86b2d41b89/helpers/transactionTypes.js#L1)

modules/clientWs/transactionSubscription.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+    });
+
+    return subscribed;
+  }
+
+  /**
+   * @returns {boolean} - whether succuessfuly subscribed to at least one transaction type
+   */
+  subscribeToTypes(...types) {
```
bludnic, line 62: Add JSDoc for `types` argument

modules/clientWs/transactionSubscription.js
```diff
@@ -0,0 +1,88 @@
...
+    return subscribed;
+  }
+
+  impliesAddress(address) {
+    return this.addresses.has(address);
+  }
+
+  impliesTransactionType(type) {
+    return this.types.has(type);
+  }
```
bludnic, line 85: Add JSDoc

test/unit/modules/clientWs.js
```diff
@@ -0,0 +1,292 @@
+'use strict';
+
+var expect = require('chai').expect;
+const TransactionType = require('../../../helpers/transactionTypes');
+
+var TransactionSubscription = require('../../../modules/clientWs/transactionSubscription');
```
bludnic, line 6: We can use `const` instead `var`

test/unit/modules/clientWs.js
```diff
@@ -0,0 +1,292 @@
...
+
+describe('TransactionSubscription', function () {
```
bludnic, line 8: Consider arrow function. Fix also below

helpers/tranasctionTypesBoundary.js
bludnic: Seems reliable. Maybe add a note that this is needed only to exclude [CHAT_MESSAGE_TYPES](https://github.com/Adamant-im/adamant/blob/fd4d79f6588e1b729859ea2bc2b8ae86b2d41b89/helpers/transactionTypes.js#L14)

bludnic (changes requested):

modules/clientWs/transactionSubscription.js
yoxira: 1. it doesn't check for a type (`string`)
2. it has the opposite condition that **allows** empty strings

bludnic: Do you know why in the validator, an empty string is a valid address?

bludnic: Maybe it's better to move additional checks in the validator?

modules/clientWs.js
```diff
@@ -65,14 +79,10 @@ function getUTime () {
...
-function findSubs (address1, address2, subs) {
-  const filterred = [];
-  for (let aId in subs) {
-    if (aId.startsWith(address1) || aId.startsWith(address2)) {
-      filterred.push(subs[aId]);
-    }
-  }
-  return filterred;
+function findSubs (transaction, subs) {
+  return subs.filter((sub) =>
```
bludnic, line 83: `TypeError: subs.filter is not a function`

bludnic (changes requested):

bludnic:
```js
import { io } from "socket.io-client";

const connection = io("http://localhost:36668", {
  reconnection: false,
  timeout: 5000,
});

connection.on("connect", () => {
  console.log("connected");

  connection.emit("address", ["U14724553101605161229"]);
  connection.emit("types", [0, 8]);
});

connection.on("newTrans", (transaction) => {
  console.log(transaction);
});
```

I run this code on my local node but cannot receive a `newTrans` event.
Tried to send an ongoing/outgoing message to/from U14724553101605161229 address.

Am I doing it right? @yoxira

bludnic (approved):

PR #82, opened 2024-09-14
Chore/update depends

yoxira:
We can't bump the `chai`, `change-case` and `execa` package versions as they dropped CommonJS support.


yoxira:
why is `package-lock.json` listed in the `.gitignore` file? 🙂

yoxira:
https://ahead.adamant.im/api/loader/status/sync

adamant-al:
> We can't bump the `chai`, `change-case` and `execa` package versions as they dropped CommonJS support.

For the next version, find alternatives and replace these.

bludnic:
> > We can't bump the `chai`, `change-case` and `execa` package versions as they dropped CommonJS support.
> 
> For the next version, find alternatives and replace these.

Let's migrate to ESM. At some point, we'll have to do it anyway. ESM can still import CommonJS, so we can at least do a partial migration.

adamant-al:
> > > We can't bump the `chai`, `change-case` and `execa` package versions as they dropped CommonJS support.
> > 
> > 
> > For the next version, find alternatives and replace these.
> 
> Let's migrate to ESM. At some point, we'll have to do it anyway. ESM can still import CommonJS, so we can at least do a partial migration.

Agreed. But also next version.

package.json
```diff
@@ -37,67 +37,67 @@
...
+    "strftime": "=0.10.3",
+    "unzipper": "^0.12.3",
     "valid-url": "=1.0.9",
     "validator.js": "=2.0.4",
-    "z-schema": "=5.0.3"
+    "z-schema": "=6.0.2"
   },
   "devDependencies": {
-    "chai": "^4.3.10",
+    "chai": "^4.1.1",
```
bludnic, line 83: Did you downgrade the version on purpose? Is it because `chai` dropped CommonJS support?

helpers/checkIpInList.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 var _ = require('lodash');
-var ip = require('ip');
+var ip = require('neoip');
```
bludnic, line 4: Why switched to `neoip`? As I see there are no differences in the libs API.

yoxira: [chore: replace node-ip with neoip to fix an npm vulnerability](https://github.com/Adamant-im/adamant/pull/82/commits/dd9fef9b0fc9c1d7b7561a20e7326aac28fccdaa)

adamant-al:
> why is `package-lock.json` listed in the `.gitignore` file? :/

Fix it

package.json
bludnic: @martiliones

bludnic:
Add `package-lock.json` to git. Use Node v18. @martiliones @yoxira

adamant-al:
Can we also update this and merge to `dev`?

test/unit/logic/account.js
```diff
@@ -0,0 +1,156 @@
+"use strict";
```
bludnic, line 1: Do you need this file in that PR? You already added it in #86

https://github.com/Adamant-im/adamant/pull/86/files#diff-a5a19004bec268761189807baa4cd27c71e35301d3f2b6dbeb9867bc82545feb

bludnic (approved):

PR #85, opened 2024-10-08
feat(WebSocket): subscribe to asset chat types

martiliones:
Automatically subscribes to tranasction type = 8 (message transaction) and watches for the subscribed `asset.chat.type`. Subscribing works the same way as for `types`.

```js
const signalMessageType = 3; 
socket.emit('assetChatTypes', signalMessageType);

socket.on('newTrans', (tx) => console.log(tx.asset.chat.type));
```

### DIFF
--- a/modules/clientWs/transactionSubscription.js
+++ b/modules/clientWs/transactionSubscription.js
@@ -2,7 +2,10 @@ const ZSchema = require('../../helpers/z_schema');
 const {
   MIN_TRANSACTION_TYPE,
   MAX_TRANSACTION_TYPE,
+  MIN_CHAT_MESSAGE_TRANSACTION_TYPE,
+  MAX_CHAT_MESSAGE_TRANSACTION_TYPE
 } = require('../../helpers/tranasctionTypesBoundary');
+const transactionTypes = require('../../helpers/transactionTypes');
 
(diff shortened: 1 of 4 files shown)


modules/clientWs/transactionSubscription.js
```diff
@@ -22,6 +25,12 @@ class TransactionSubscription {
      * @type {Set<number>}
      */
     this.types = new Set();
+
+    /**
+     * List of asset chat types to subscribe to
+     * @type {Set<number>}
+     */
+    this.assetChatTypes = new Set();
```
bludnic, line 33: Suggestion: Since we don't have such a thing `asset.type` and only `asset.chat.type`, consider using `assetTypes` naming. Check other places as well.

Resolve if you leave as is. @martiliones

bludnic: Approved in advance

bludnic (approved):

bludnic:
Is there a Wiki on how to use connect to the node using WebSockets? Maybe we should add it? @yoxira 
https://github.com/Adamant-im/adamant/wiki

adamant-al:
@yoxira 
Update

yoxira:
https://github.com/Adamant-im/adamant/wiki/REST-and-Socket-connections

PR #86, opened 2024-11-07
Chore/tests

yoxira:


test/unit/logic/state.js
```diff
@@ -0,0 +1,362 @@
...
+    });
+  });
+
+  describe('dbSave', () => {
+    it('should throw an error with no param', () => {
+      expect(state.dbSave).to.throw();
+    });
+
+    it('should return promise object for valid parameters', () => {
+      const savePromise = state.dbSave(validTransaction);
```
bludnic, line 337: Better to avoid such naming, because the return type is no a promise

test/unit/logic/chat.js
```diff
@@ -0,0 +1,581 @@
...
+  timestamp: 63137661,
+  signature:
+    'e25f1aba994c7f07c03099edcbe0ada19df371ddf1a829dae8dee36ab809ce8a438111bf65056c813e9dc832a890a081ba1cd295d37e509f62f042149e62e30d',
+  id: '8958126469643732641',
+  fee: 100000,
+  relays: 1,
+  receivedAt: '2019-09-03T11:14:22.638Z',
+};
+
+describe('Chat', () => {
```
bludnic, line 105: There are failed tests

image

test/unit/logic/chat.js
```diff
@@ -0,0 +1,581 @@
...
+
+      chat.verify(trs, validSender, function (err) {
+        expect(err).to.equal(
+          'Message is too long. Maximum is 20480 characters'
+        );
+        done();
+      });
+    });
+
+    it('should verify okay for valid transaction', (done) => {
```
bludnic, line 347:
```suggestion
    it('should verify for a valid transaction', (done) => {
```

test/unit/modules/delegates.js
```diff
@@ -0,0 +1,379 @@
...
+const { expect } = require('chai');
+
+const { modulesLoader } = require('../../common/initModule.js');
+const {
+  testAccount,
+  invalidPublicKey,
+} = require('../../common/stubs/account.js');
+
+const constants = require('../../../helpers/constants.js');
+const Delegates = require('../../../modules/delegates.js');
```
bludnic, line 12: This import is not used

test/unit/modules/transactions.js
```diff
@@ -0,0 +1,903 @@
+'use strict';
+
+const { expect } = require('chai');
+
+const Transactions = require('../../../modules/transactions.js');
```
bludnic, line 5: Import not used

test/unit/modules/transactions.js
```diff
@@ -0,0 +1,903 @@
...
+const {
+  unconfirmedTransaction,
+  nonExistingTransactionId,
+  unconfirmedTransactionId,
+  existingTransaction,
+  existingTransactionWithAsset,
+} = require('../../common/stubs/transactions.js');
+const { genesisBlockId } = require('../../common/stubs/blocks.js');
+
+describe('transactions', function () {
```
bludnic, line 35: Some tests failed

image

test/unit/modules/accounts.js
```diff
@@ -0,0 +1,666 @@
...
+
+    describe('open()', () => {
+      it('should return error if "secret" is of invalid type', (done) => {
+        const body = { secret: true };
+        accounts.shared.open({ body }, (err, response) => {
+          expect(response).not.to.exist;
+          expect(err).to.include('Expected type string but found');
+          done();
+        });
+      });
```
bludnic, line 355: Since you added an additional check in `openAccount` method:

https://github.com/Adamant-im/adamant/blob/d493f17446d8ad835c1a419b83daead72a3367db/modules/accounts.js#L97-L101

Add also a test case for `{ secret: "invalid secret"  }`. It must throw "Mnemonic string is invalid:..."

bludnic (changes requested):

test/unit/modules/delegates.js
yoxira: it's used for type declaration in jsdoc
```js
describe('delegates', function () {
  /**
   * @type {Delegates}
   */
  let delegates;
```

bludnic (approved):

yoxira:
Not ready for a review.

yoxira:
@adamant-al I think that tests under `test/api` replace the need for "transaction flood script" (https://trello.com/c/T1RLgwVF), as they include stress tests for the transaction types. We will run the `test/api` tests on every new version, and the `test/unit` tests, which don’t require running a local test node, are there for general development.

yoxira:
Note that the PR aims to fix previous tests, add more unit tests, and resolve issues caused by updating dependencies (#82). Some tests, like `test/api/peer.transactions.votes.js`, may still fail due to unexpected node behavior, _not the mistakes in tests_.

bludnic:
@yoxira why downgrading `change-case` and `execa` versions? https://github.com/Adamant-im/adamant/pull/86/commits/53c0bc9c198244ae66ee6252a2e1dba69fe6f875

modules/accounts.js
```diff
@@ -154,7 +154,14 @@ Accounts.prototype.generateAddressByPublicKey = function (publicKey) {
...
 Accounts.prototype.getAccount = function (filter, fields, cb) {
   if (filter.publicKey) {
-    filter.address = self.generateAddressByPublicKey(filter.publicKey);
+    try {
+      filter.address = self.generateAddressByPublicKey(filter.publicKey);
+    } catch (error) {
+      if (typeof fields === 'function') {
+        return setImmediate(fields, error);
+      }
+      return setImmediate(cb, error);
```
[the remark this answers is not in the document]
bludnic: Cover with tests [test/unit/modules/accounts.js](https://github.com/Adamant-im/adamant/blob/82771c92054574922281740e83666fc535afe0c6/test/unit/modules/accounts.js) different scenarios:

1. With fields: 
```js
account.getAccount({ publicKey }, ['balance'], (err) => {})
```
2. Without fields
```js
account.getAccount({ publicKey }, (err) => {})
```
3. Edge cases (e.g. `cb` not provided, or `fields` as non-array)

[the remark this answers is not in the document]
bludnic: Update also the JSDoc describing different function signatures.
https://austingil.com/typescript-function-overloads-with-jsdoc/

bludnic:
@yoxira Well done described tests and best practices in [CONTRIBUTING.md](https://github.com/Adamant-im/adamant/pull/86/commits/a3262e5cde357b892643d4c2a33abbad75fffd4b#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44c) 👍

package.json
```diff
@@ -80,7 +80,7 @@
     "z-schema": "=6.0.2"
   },
   "devDependencies": {
-    "chai": "^5.1.2",
+    "chai": "^4.3.10",
```
bludnic, line 84: Can you describe the issue with the newer version of `chai`?

helpers/sequence.js
```diff
@@ -18,15 +18,20 @@ function Sequence (config) {
...
     if (_default.onWarning && self.sequence.length >= _default.warningLimit) {
       _default.onWarning(self.sequence.length, _default.warningLimit);
     }
-    self.__tick(function () {
-      setTimeout(nextSequenceTick, 3);
-    });
-  });
+
+    self.__tick(self.nextSequenceTick);
+  };
```
bludnic, line 34: Can you explain why this change? Does it fix a bug? @yoxira 

1. I don't see `nextSequenceTick` to be initially invoked on first run anymore (previously via `setImmediate()`).
2. I think there was a reason in using `setTimeout` (for non-blocking). Without it, it may result in stack overflow.
3. Tests needed.

Better to revert the changes if it does not fix any bug.

modules/cache.js
```diff
@@ -44,18 +44,21 @@ Cache.prototype.isReady = function () {
...
+  let parsedValue;
+
+  try {
+    const value = await client.get(key);
+    parsedValue = JSON.parse(value);
+  } catch (err) {
+    cb(err, key);
+  }
+
+  cb(null, parsedValue);
```
bludnic, line 61: You call the `cb()` twice when an error is caught. Is that expected?

modules/delegates.js
```diff
@@ -870,7 +870,7 @@ Delegates.prototype.shared = {
 
         modules.accounts.getAccounts({
           address: { $in: addresses },
-          sort: 'balance'
+          sort: { balance: -1 }
```
bludnic, line 873: Add unit tests for this case

package.json
```diff
@@ -77,6 +77,7 @@
     "unzipper": "^0.12.3",
     "valid-url": "=1.0.9",
     "validator.js": "=2.0.4",
+    "why-is-node-running": "^2.3.0",
```
bludnic, line 80: Forgot to remove?

bludnic (changes requested):

yoxira:
@bludnic I accidentally bumped them in #82, even though I explicitly wrote the description in the PR:
> We can't bump the chai, change-case and execa package versions as they dropped CommonJS support.

helpers/sequence.js
yoxira: @bludnic Previously, the tick function ran indefinitely every `N` ms, preventing the process from exiting when the script ended (e.g., after all unit tests passed).

1. It now starts only when a new task is added to the queue, instead of continuously checking for new tasks.
2. It still avoids blocking the event loop, thanks to `setImmediate`.

bludnic: I'm wondering if there was a reason for using `setTimeout` + `setImmediate` in previous version. Currently, only `setImmediate` is used. Also, the `setTimeout` had a delay of 3ms previously.

Anyway, since you touched that file, better to cover it with tests.

bludnic: Consider whether there may be a scenario in which the queue suddenly stops.

test/unit/modules/transactions.js
bludnic: two tests failed

image

bludnic (changes requested):

modules/accounts.js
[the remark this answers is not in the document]
yoxira: The edge cases are not that important because it's an internal method

modules/accounts.js
```diff
@@ -143,14 +147,36 @@ Accounts.prototype.generateAddressByPublicKey = function (publicKey) {
...
- * @param {function} fields - Fields to get.
+ * @param {Array} fields - Fields to get.
  * @param {function} cb - Callback function.
  */
+
+/**
+ * Gets account information, calls logic.account.get().
+ * @overload
+ * @param {Object} filter - Contains publicKey.
+ * @param {function} fields - Callback function.
```
bludnic, line 160:
```suggestion
 * @param {function} cb - Callback function.
```

bludnic (approved):

PR #89, opened 2024-12-16
feat: disconnect from peers on connection errors 

yoxira:
https://trello.com/c/l0PvLgan/97-check-adding-and-removing-new-peers

- Made removing and adding a new peer the same log level (`info`)
- Instead of removing a peer, multiple failed requests will change a peer state to `DISCONNECT`
  
  > Previously, the connection errors behavior was to remove the peer from the peers list. However, if the error happens due to a request-limiter policy or internet issues, the node only removes the peer temporarily until it rediscovers it in a second after removing, leading to spam in logs.


yoxira:
To do: move request success rate and request sample to a config file

adamant-al:
image

Firstly, init nodes added, and then others discovered.

```
http://162.55.32.80:36666/api/peers
```

Summary:
	•	Total Peers Count: 137
	•	Peers by Version:
	•	0.8.3: 95
	•	0.8.1: 20
	•	0.8.0: 10
	•	0.7.0: 2
	•	0.7.1: 1
	•	0.6.5: 9

Peers by State:
	•	Active (state: 2): 133
	•	Inactive (state: 1): 4

logic/peer.js
```diff
@@ -72,13 +74,48 @@ Peer.prototype.nullable = [
...
 };
 
 // Public methods
+/**
+ * Calculate requests success rate.
+ * @returns {number} Success percentage
+ */
+Peer.prototype.calcSuccessRate = function () {
+  const successRate = (this.successRequestCount / SUCCESS_RATE_POOL_SIZE) * 100;
+  return Math.round(successRate * 100) / 100;
```
bludnic, line 95: Why not using just: `return Math.round(successRate)`?

modules/peers.js
```diff
@@ -299,6 +299,16 @@ Peers.prototype.update = function (peer) {
   return library.logic.peers.upsert(peer);
 };
 
+Peers.prototype.isFrozen = function (ip, port) {
```
bludnic, line 308: Add JSDoc

modules/transport.js
```diff
@@ -98,6 +98,21 @@ __private.removePeer = function (options, extraMessage) {
   modules.peers.remove(options.peer.ip, options.peer.port);
 };
 
+/**
+ * Updates the request success rate for the peer.
+ * Provide the error if the request has been failed.
+ * @implements modules.peers.recordRequest
+ * @param {{ip: string, port: number, error?: string}} options
```
bludnic, line 105:
```suggestion
 * @param {{ peer: { ip: string, port: number }, error?: string }} options
```

bludnic (changes requested):

modules/peers.js
yoxira: it's there..?

https://github.com/Adamant-im/adamant/blob/feat/peers/modules/peers.js#L308

logic/peer.js
yoxira: the rounding here is unnecessary, it's just for logs if we ever need one, and decimal numbers with double digits after the point just seem fancier

bludnic (approved):

PR #91, opened 2024-12-17
feat: add `includeDirectTrasnfers` param for the `chats/get` endpoint

### LINKED ISSUE #170: [Enhancement] Add includeDirectTransfers parameter to /api/chats/get
### Summary

Add a positive `includeDirectTransfers` query parameter to `/api/chats/get`, keep backward compatibility with the deprecated `withoutDirectTransfers` parameter, and align the endpoint behavior with the existing chat filtering model.

### Motivation

The existing `withoutDirectTransfers` filter is hard to reason about because it uses a double negative. A positive parameter name is clearer for API consumers, reduces integration mistakes, and makes `/api/chats/get` consistent with related chat endpoints and query-language documentation.

### Detailed description

This issue is created as the tracking issue for work that was implemented in [PR #91](https://github.com/Adamant-im/adamant/pull/91), which was merged on February 25, 2025.

Scope implemented in that PR:

- add `includeDirectTransfers` support for `/api/chats/get`
- preserve backward compatibility by continuing to accept deprecated `withoutDirectTransfers`
- keep `withoutDirectTransfers` as the higher-priority option when both parameters are provided
- cover the behavior with API tests for default behavior and explicit parameter values

Relevant node changes:

- `modules/chats.js`
- `test/api/chats.js`

Companion artifacts are already aligned at the time of issue creation:

- `Adamant-im/adamant-schema` documents `includeDirectTransfers` and marks `withoutDirectTransfers` as deprecated
- `Adamant-im/docs` documents `includeDirectTransfers` for `/api/chats/get` and related chat endpoints

### Screenshots or videos

N/A

### Alternatives

Keep only `withoutDirectTransfers` and document the double negative more clearly. This would preserve compatibility but would keep the API harder to use and less consistent.

Another alternative would be a breaking rename that removes `withoutDirectTransfers` immediately, but that would create avoidable integration risk for existing clients.

- add `includeDirectTransfers` parsing in `/api/chats/get`
- keep backward-compatible handling of `withoutDirectTransfers`
- prefer `withoutDirectTransfers` when both filters are present
- verify behavior with API tests
- keep schema/docs synchronized with the node implementation

yoxira:
https://trello.com/c/gX2i1brG/115-feat-add-withoutdirecttransfers-option-to-chats-get

title

### DIFF
--- a/modules/chatrooms.js
+++ b/modules/chatrooms.js
@@ -67,7 +67,18 @@ __private.listChats = function (filter, cb) {
     where.push('"c_type" = ${type}');
     params.type = filter.type;
   }
-  if (filter.withoutDirectTransfers) {
+
+  let includeDirectTransfers = false;
+
+  if (typeof filter.includeDirectTransfers !== 'undefined') {
+    includeDirectTransfers = filter.includeDirectTransfers;
+  }
+
+  if (typeof filter.withoutDirectTransfers !== 'undefined') {
+    includeDirectTransfers = !filter.withoutDirectTransfers;
+  }
+
+  if (includeDirectTransfers) {
     where.push('"t_type" = ' + transactionTypes.CHAT_MESSAGE);
   } else {
     where.push(`("t_type" = ${transactionTypes.CHAT_MESSAGE} OR "t_type" = ${transactionTypes.SEND})`);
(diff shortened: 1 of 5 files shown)


yoxira:
To do: tests

yoxira:
@adamant-al 
Setting `?withoutDirectTransfers` to `false` by default is very confusing due to double negative, can we rename the property name? e.g., to `includeDirectTransfers`

adamant-al:
> @adamant-al Setting `?withoutDirectTransfers` to `false` by default is very confusing due to double negative, can we rename the property name? e.g., to `includeDirectTransfers`

Yes, but let's add `includeDirectTransfers` where it's currently `withoutDirectTransfers`
> It’s supported by the api/chatrooms endpoint (here default withoutDirectTransfers=0.

And deprecate `withoutDirectTransfers`.
If both options included, `withoutDirectTransfers` is a priority.

bludnic:
@yoxira any update on this?

yoxira:
To Do: update wiki

Another reason why we should host our own documentation site is having a separate dev branch with unreleased features.

modules/chatrooms.js
```diff
@@ -164,10 +175,21 @@ __private.listMessages = function (filter, cb) {
...
     where.push(`(NOT("c_type" = ${transactionTypes.CHAT_MESSAGE_TYPES.SIGNAL_MESSAGE}) OR c_type IS NULL)`);
   }
-  if (filter.withoutDirectTransfers) {
-    where.push('"t_type" = ' + transactionTypes.CHAT_MESSAGE);
-  } else {
+
+  let includeDirectTransfers = false;
+
+  if (typeof filter.includeDirectTransfers !== 'undefined') {
+    includeDirectTransfers = filter.includeDirectTransfers;
```
bludnic, line 182: Maybe safer to convert to boolean?

yoxira: the variable is only used in an "if" statement so there's no need to convert, but I think it's good for convention

bludnic (approved):

metalisk:
Backfilled tracking issue: #170

This feature request is now tracked in issue #170 and was implemented in this PR.

PR #93, opened 2024-12-17
feat: add `timestampMs` for transactions

### LINKED ISSUE #209: [Feat] Verify, store and return timestampMs for transactions
## Summary

Add optional `timestampMs` support to ADAMANT transactions so clients can preserve millisecond-level ordering for fast message flows without changing transaction signatures, transaction IDs, or existing `timestamp` semantics.

## Motivation

Transactions currently include `timestamp` as ADAMANT epoch time in seconds. This field is part of transaction signature and ID calculation, so a node must not rewrite it after receiving a signed transaction from a client.

In fast chats, several message transactions can share the same second-level `timestamp`, which can make message order unstable for clients. A soft, backward-compatible `timestampMs` field gives clients more precise ordering while preserving consensus safety before an explicit protocol activation.

## Detailed description

The node should accept, store, validate, and return an optional `timestampMs` transaction field with the following behavior:

- Keep `timestamp` as the consensus-critical ADAMANT epoch timestamp in seconds.
- Add `timestampMs` as ADAMANT epoch time in milliseconds, not Unix milliseconds.
- Clients should derive `timestamp` from the same millisecond source using `Math.floor(timestampMs / 1000)`; `round` and `ceil` can create an invalid pair.
- Do not include `timestampMs` in transaction bytes, signatures, second signatures, or transaction IDs.
- Treat `timestampMs` as optional for compatibility with existing clients and historical transactions.
- When a client posts a new transaction with `timestampMs`, verify that it belongs to the same ADAMANT epoch second as `timestamp`: `0 <= timestampMs - timestamp * 1000 < maxTimestampMsDelta`.
- Reject invalid `timestampMs` values with a clear API error.
- Allow a small public-API future-time grace for fresh transactions via `maxTransactionFutureMs = 500`; this is admission policy only and is not consensus/replay validation.
- Store `timestampMs` in `trs` and expose it through transaction read models.
- Return `timestampMs` in API responses and socket/WebSocket transaction payloads wherever transaction objects are returned.
- Prefer `timestampMs` for timestamp-based transaction ordering when it is available, with `timestamp` as the fallback.

The implementation must preserve backward compatibility:

- Before the `spaceship` activation height, nodes must not reject or persist peer-propagated transactions only because they include an invalid or unsupported `timestampMs`.
- After `spaceship` activation, nodes may keep and validate `timestampMs` during transaction normalization and verification.
- Existing transactions and blocks without `timestampMs` must remain valid.
- Chain replay from genesis must remain deterministic.

Related pull requests:

- https://github.com/Adamant-im/adamant/pull/93
- https://github.com/Adamant-im/adamant/pull/210

Related AIP:

- https://github.com/Adamant-im/AIPs/pull/62

Companion documentation/schema PRs:

- https://github.com/Adamant-im/adamant-schema/pull/27
- https://github.com/Adamant-im/docs/pull/21

Related sorting follow-up:

- #182

## Alternatives

- Keep only second-level `timestamp` and let clients guess ordering. This preserves current behavior but does not solve fast message ordering.
- Let clients subtract a default time correction before signing transactions. This helps the "timestamp is in the future" edge case but does not provide stable millisecond ordering.
- Change transaction serialization to include milliseconds. This would be a consensus and wire-format change, requiring a full protocol upgrade and coordinated client/node rollout.

- Add `maxTimestampMsDelta`, `maxTransactionFutureMs`, and public-API freshness constants in `helpers/constants.js`.
- Add `slots.getTimeMs()` while keeping `slots.getTime()` in seconds.
- Add a `logic/consensus` activation helper with a `spaceship` activation height.
- Add `timestampMs` to transaction creation, schema validation, DB fields, DB save/read, and SQL migrations.
- Gate peer/block normalization so `timestampMs` is removed before `spaceship` activation and kept after activation.
- Validate the `timestampMs` delta only when the field survives normalization.
- Keep public API freshness checks (`future timestamp`, too-old transaction window) in `publish()` because those checks depend on current node time and must not be applied to historical replay. The `maxTransactionFutureMs` grace only softens the next-slot boundary case for fresh public API submissions.
- Ensure SQL projections include `t_timestampMs` for list, full-list, get-by-id, chat, state, and any socket response paths that serialize transactions.
- Ensure timestamp ordering paths use `timestampMs` first and `timestamp` as fallback where the API exposes `orderBy=timestamp`.
- Return `nodeTimestampMs` and `unixTimestampMs` from `/api/node/status` as non-consensus node clock metadata.
- Keep consensus activation heights configurable for testnet and ADAMANT-based chains while preserving mainnet defaults.
- Treat `fairSystem` as a named activation with first active height `4,359,465`, equivalent to the legacy `height > fairSystemActivateBlock` behavior.

## Consensus and activation notes

The `spaceship` activation should cover behavior that can make two nodes disagree about transaction acceptance:

- preserving `timestampMs` during peer/block normalization;
- validating `timestampMs` during synchronization, block processing, and peer transaction processing;
- any future rule that would make a transaction invalid because of `timestampMs`.

The activation should not cover purely local/public-API checks that are only applied before the first node accepts a freshly submitted transaction, such as "timestamp is too far in the past" based on current node time. Those checks are not replay-stable and should stay outside consensus verification.

The activation boundary must be tested with the height being validated, not only the node's current loader height, so the exact activation block behaves deterministically during live processing, synchronization, rollback, and replay.

Clock-relative public API admission checks, including the `maxTransactionFutureMs = 500` future timestamp grace for fresh transactions, are not consensus behavior and should not be gated by `spaceship`. A transaction timestamp can be slightly ahead of the block timestamp without causing consensus divergence because transaction time is signed metadata, not the source of block-slot validity.

## Checklist

- [ ] Confirm the exact `spaceship` activation height before release.
- [x] Verify that `timestampMs` is not part of transaction bytes, signatures, second signatures, or transaction IDs. Covered by #93.
- [x] Add `timestampMs` to transaction creation without changing existing `timestamp` semantics. Covered by #93.
- [x] Add DB migration, index, and all affected transaction views/projections. Base DB/index covered by #93; remaining response views/projections covered by #210.
- [x] Store and read `timestampMs` through transaction DB save/read paths. Covered by #93; numeric parsing hardened in #210.
- [x] Return `timestampMs` through transaction API responses, chat/state transaction responses, and socket/WebSocket payloads. Remaining SQL projection gaps covered by #210.
- [x] Remove `timestampMs` before activation in consensus-sensitive normalization. Covered by #93.
- [x] Keep and validate `timestampMs` after activation. Covered by #93 after `spaceship`; activation-height context tightened in #210.
- [x] Keep fresh public-API timestamp window checks in `publish()` only. Covered by #93 and retained in #210; `maxTransactionFutureMs = 500` grace is covered by #210.
- [x] Add pre-activation and post-activation tests around `objectNormalize()` and `verify()`. Covered by #93; explicit block-height activation test added in #210.
- [ ] Add API/list response tests proving `timestampMs` is returned where expected.
- [x] Add ordering tests proving `timestampMs` is preferred and `timestamp` remains the fallback. Covered by #210.
- [x] Add tests proving `timestampMs` does not affect signatures, transaction IDs, or transaction hashes. Covered by #93 and expanded in the follow-up branch.
- [x] Add tests proving invalid `timestampMs` is ignored before `spaceship` and rejected after it. Covered by #93/#210 and expanded in the follow-up branch.
- [x] Add `/api/node/status` unit coverage for `nodeTimestampMs` and `unixTimestampMs`. Covered by the follow-up branch.
- [x] Add configurable activation-height coverage for `spaceship` and `fairSystem`. Covered by the follow-up branch.
- [x] Update companion schema/docs repositories if API response shape changes. Covered by Adamant-im/adamant-schema#27 and Adamant-im/docs#21.

## Verification

The task is complete when:

- existing pre-activation blocks and transactions replay successfully;
- transactions without `timestampMs` remain valid;
- transactions with valid `timestampMs` are stored and returned after activation;
- transactions with invalid `timestampMs` are rejected after activation in consensus-sensitive paths;
- public API rejects future and too-old fresh transactions without applying those current-time checks to historical replay;
- transaction lists, chat-related responses, and socket payloads include `timestampMs` when it is stored;
- tests cover both sides of the `spaceship` activation boundary;
- targeted unit tests, relevant API tests, migration checks, and lint are reported in the implementing PR.

yoxira:
https://trello.com/c/IN62zyEb/114-feat-verify-store-and-return-unixtimestamp-for-txs

Add the **optional** `timestampMs` field for more precise order of messages.

>[!NOTE]
> `0.8.4-dev.0` version can still not be accepted by the old test nodes, you should change it to `v0.8.4` or similar to avoid `Inadequate broadhash consensus` error.


logic/transaction.js
```diff
@@ -476,6 +479,10 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
     return setImmediate(cb, 'Missing sender');
   }
 
+  if (typeof trs.timestampMs !== 'number') {
```
adamant-al, line 482: Will previous txs/blocks be valid?
It's important to keep backward compatibility.

logic/transaction.js
```diff
@@ -647,10 +654,26 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
   if (trs.timestamp < INT_32_MIN || trs.timestamp > INT_32_MAX) {
     return setImmediate(cb, 'Invalid transaction timestamp. Timestamp is not in the int32 range');
   }
-  if (slots.getSlotNumber(trs.timestamp) > slots.getSlotNumber()) {
+
+  if (Math.abs(trs.timestampMs - trs.timestamp * 1000) >= 1000) {
```
adamant-al, line 658: Will previous txs/blocks be valid?

logic/transaction.js
```diff
@@ -647,10 +654,26 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+    return setImmediate(cb, 'Invalid transaction timestamp. Timestamp and timestampMs delta is greater than 1000ms');
+  }
+
+  const currentTime = slots.getTime();
```
adamant-al, line 662: Returns time in seconds, not in ms, right?

adamant-al:
Ensure that

> A node will include `timestampMs` in responses, including socket.

logic/transaction.js
yoxira: yes

yoxira:
Test script using `adamant-api`:


```js
import {
  createSendTransaction,
  fees,
  createKeypairFromPassphrase,
  AdamantApi,
} from 'adamant-api';

const transaction = createSendTransaction({
  recipientId: RECIPIENT_ADDRESS,
  amount: fees.send,
  keyPair: createKeypairFromPassphrase(SENDER_PASSPHRASE),
});

transaction.timestampMs = transaction.timestamp * 1000;

const res = await api.post('transactions/process', {
  transaction,
});

console.log(res);
```

yoxira:
Now `timestampMs` isn't mandatory.

- `Transaction.prototype.verify` - checks a transaction retrieved from other blocks; may include old transactions.
- `Transactions.prototype.publish` - checks incoming transactions from users, verifying "fresh" unconfirmed transactions.

logic/transaction.js
```diff
@@ -662,6 +683,58 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+  const { timestamp, timestampMs } = trs;
+
+  if (timestamp < INT_32_MIN || timestamp > INT_32_MAX) {
+    return 'Invalid transaction timestamp. Timestamp is not within the int32 range';
+  }
+
+  if (typeof timestampMs === 'number') {
+    const timestampMsDelta = Math.abs(timestampMs - timestamp * 1000);
+
+    if (timestampMsDelta >= 1000) {
```
bludnic, line 691: 1000 -> const

logic/transaction.js
```diff
@@ -481,6 +491,18 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+    return setImmediate(cb, 'Invalid transaction timestamp. Timestamp is not within the int32 range');
+  }
+
+  if (typeof timestampMs === 'number') {
+    const timestampMsDelta = Math.abs(timestampMs - timestamp * 1000);
+
+    if (timestampMsDelta >= 1000) {
+      return setImmediate(cb, 'Invalid transaction timestamp. The difference between timestamp and timestampMs is greater than 1000ms');
+    }
+  }
```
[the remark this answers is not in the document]
yoxira: the code was moved to `validateTimestampMs()`.

logic/transaction.js
```diff
@@ -662,6 +669,63 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+
+    const { maxTimestampMsDelta } = constants;
+    if (timestampMsDelta >= maxTimestampMsDelta) {
+      return `Invalid transaction timestamp. The difference between timestamp and timestampMs is greater than ${maxTimestampMsDelta}ms`;
+    }
+  }
+};
+
+/**
+ * Validates the timestamp and timestampMs of a transaction to ensure they are within an acceptable range
```
adamantmm, line 699: If it's used for new txs only and not used for past txs validation (doesn't influence consensus), consider adding such a comment.

adamantmm (changes requested):

adamantmm:
> Now `timestampMs` isn't mandatory.
> 
> * `Transaction.prototype.verify` - checks a transaction retrieved from other blocks; may include old transactions.
> * `Transactions.prototype.publish` - checks incoming transactions from users, verifying "fresh" unconfirmed transactions.

Can you confirm that?

logic/transaction.js
```diff
@@ -120,6 +125,12 @@ Transaction.prototype.publish = function (data) {
     throw 'Invalid signature';
   }
 
+  const error = this.verifyTimestamp(data);
```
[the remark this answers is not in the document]
bludnic: It's a string containing the error message, or `undefined` if validation passed. A better name, in my opinion, would be `errorMessage`.

logic/transaction.js
```diff
@@ -662,6 +669,63 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
 
+/**
+ * Validates timestamp and timestampMs formats
+ * @param {Transaction} trs - The transaction object to validate
+ * @returns {string | undefined} - Returns string if the timestamp is invalid with the provided reason
+ */
+Transaction.prototype.validateTimestampMs = function (trs) {
+  const { timestamp, timestampMs } = trs;
+
+  if (timestamp > INT_32_MAX) {
```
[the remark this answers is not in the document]
yoxira: in 2087

The check for `int_32_max` was always there fyi https://github.com/Adamant-im/adamant/blob/master/logic/transaction.js#L647

logic/transaction.js
```diff
@@ -662,6 +669,63 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+    return 'Invalid transaction timestamp. Timestamp is not within the int32 range';
+  }
+
+  if (timestamp < 0) {
+    return 'Invalid transaction timestamp. The timestamp is after the epoch time'
```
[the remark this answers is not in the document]
yoxira: I removed it completely to avoid a possible consensus failure.

yoxira:
> Can you confirm that?

To be precise, `verify` is also called after receiving new transaction from the API, following `publish` and `objectNormalize` methods. And `publish` is only called when pushing transactions via the public API.

yoxira:
After changes:

 - Removes `timestampMs` from transactions before reaching a consensus. Important: the property is also removed before broadcasting to other nodes
 - After meeting a consensus, validates and includes `timestampMs` in the transactions
- Regardless of a consensus, checks if the transaction retrieved from the public API has a `timestamp` that is not in the future nor too far in the past

yoxira:
I suggest to use consensus code name in versions, e.g. `0.9.11-spaceship`

helpers/slots.js
```diff
@@ -50,6 +50,15 @@ module.exports = {
    * @return {number} ADAMANT epoch time constant.
```
adamant-al, line 50: * @return {number} ADAMANT epoch time (passed since ADM blockchain creation) constant in sec.

Add also method and parameter description to JSDoc.

logic/consensus/activationHeights.js
```diff
@@ -0,0 +1,7 @@
+const consensusActivationHeights = {
```
adamant-al, line 1: Add comments with descriptions

adamant-al:
> > Can you confirm that?
> 
> To be precise, `verify` is also called after receiving new transaction from the API, following `publish` and `objectNormalize` methods. And `publish` is only called when pushing transactions via the public API.

So what's the exact flow? E.g.,

Receiving a new tx from a client app via API (it's the first node which sees the tx):
1. one
2. two
3. ..

Receiving a new tx (current height) from another peer:
1. one
2. two
3. ..

Fetching a tx from another peer when arising/verifying blockchain from 0 height:
1. one
2. two
3. ..

Please add trace logging to each involved methods, include tx id, block height, if it's already included in block or not, what we are doing, and other reasonable info.

Maybe you also can add more tests.

Then you can run to verify it and check logs using testnet.

**It will help us to understand the process better and keep concensus.**

logic/transaction.js
```diff
@@ -120,6 +128,12 @@ Transaction.prototype.publish = function (data) {
     throw 'Invalid signature';
   }
 
+  const timestampValidationError = this.validateTimestampWindow(data);
```
adamant-al, line 131: Add JSDoc to `Transaction.prototype.publish`
Include when it's called exactly.

Add comment when you call `validateTimestampWindow()`, what's the goal and consensus name when activated

logic/transaction.js
```diff
@@ -662,6 +672,54 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
   });
 };
 
+/**
+ * Validates timestamp and timestampMs formats
+ * @param {Transaction} trs - The normalized transaction object to validate
+ * @returns {string | undefined} - Returns string if the timestamp is invalid with the provided reason
+ */
+Transaction.prototype.validateTimestampMs = function (trs) {
```
adamant-al, line 680: I'm still confused to have both `validateTimestampWindow()` and `validateTimestampWindow()`.

Consider creating a single function, you can use params what to check and what to skip.

adamant-al:
> I suggest to use consensus code name for versioning, e.g. `0.9.11-spaceship`

Are speaking about package.json? Doesn't it break something?

adamant-al:
> After changes:
> 
> * Removes `timestampMs` from transactions before reaching a consensus. Important: the property is also removed before broadcasting to other nodes
> * After meeting a consensus, validates and includes `timestampMs` in the transactions
> * Regardless of a consensus, checks if the transaction retrieved from the public API has a `timestamp` that is not in the future nor too far in the past

My guess was to keep `timestampMs`, and benefit from having it even before new consensus arrives:

- When a new node accepts the tx with `timestampMs` from a client via API (`publish`), it can check/validate it, and decline, don't even broadcast it to other nodes. Is it how it's working now, broadcasting only valid txs?
- Skip validating `timestampMs` before `spaceship` activated when receiving a tx from another node (`verify`)
- Validate `timestampMs` everywhere after `spaceship` activated

Will it work?

logic/transaction.js
```diff
@@ -1172,6 +1236,10 @@ Transaction.prototype.objectNormalize = function (trs) {
     }
   }
 
+  if (!this.scope.consensus.isActivated('spaceship')) {
```
adamant-al, line 1239: See my comment about keeping `timestampMs`
https://github.com/Adamant-im/adamant/pull/93#issuecomment-2681306317

adamant-al (changes requested):

logic/transaction.js
```diff
@@ -662,6 +672,54 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+ */
+Transaction.prototype.validateTimestampWindow = function (trs) {
+  const { timestamp } = trs;
+
+  const currentTime = slots.getTime();
+
+  const currentSlotNumber = slots.getSlotNumber(currentTime);
+  const transactionSlotNumber = slots.getSlotNumber(timestamp);
+
+  if (transactionSlotNumber > currentSlotNumber) {
```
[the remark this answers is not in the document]
yoxira: > Is it a replacement for

yes

> Is slots.getSlotNumber() the same as slots.getSlotNumber(currentTime)?

yes, `currentTime` is reused so it's a separate variable to avoid calling `getTime()` twice

logic/transaction.js
```diff
@@ -662,6 +672,54 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
...
+    return 'Transaction timestamp is in the future';
+  }
+
+  const earliestValidTime = currentTime - constants.maxTransactionAge;
+  const earliestValidSlotNumber = slots.getSlotNumber(earliestValidTime);
+
+  if (transactionSlotNumber < earliestValidSlotNumber) {
```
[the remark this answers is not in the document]
yoxira: No, this check is only for transactions from the public API.

```js
  if (transactionSlotNumber < earliestValidSlotNumber) {
    throw `Transaction timestamp is more than ... seconds in the past`;
  }
```

logic/transaction.js
```diff
@@ -481,6 +493,12 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
     return setImmediate(cb, 'Unknown transaction type ' + trs.type);
   }
 
+  const timestampValidationError = this.validateTimestampMs(trs);
```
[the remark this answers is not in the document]
yoxira: No, it won’t affect consensus. I just accidentally changed the order of possible errors. Although, it makes sense to prioritize checks that require less resources

logic/transaction.js
yoxira: I removed both functions

logic/transaction.js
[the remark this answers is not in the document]
adamant-al: Let's move it back to have less changes?
I think we'll not save much on prioritizing.

logic/consensus/activationHeights.js
```diff
@@ -0,0 +1,7 @@
...
+  spaceship: 1_000_000,
```
adamant-al:
```suggestion
  spaceship: 100_000_000,
```

We already have 45M blocks :)

yoxira:
Currently:
- `timestampMs` will be activated at height 100,000,000 ("spaceship" consensus), including validation and database storage
- Transactions received from the public API must not be older than 5 seconds or set in the future. This rule applies before the consensus activation

adamantmm (approved):

adamant-al:
@bludnic 
Please check this PR.

adamant-al:
> > I suggest to use consensus code name for versioning, e.g. `0.9.11-spaceship`
> 
> Are speaking about package.json? Doesn't it break something?

@yoxira

adamant-al:
> > > Can you confirm that?
> > 
> > 
> > To be precise, `verify` is also called after receiving new transaction from the API, following `publish` and `objectNormalize` methods. And `publish` is only called when pushing transactions via the public API.
> 
> So what's the exact flow? E.g.,
> 
> Receiving a new tx from a client app via API (it's the first node which sees the tx):
> 
> 1. one
> 2. two
> 3. ..
> 
> Receiving a new tx (current height) from another peer:
> 
> 1. one
> 2. two
> 3. ..
> 
> Fetching a tx from another peer when arising/verifying blockchain from 0 height:
> 
> 1. one
> 2. two
> 3. ..
> 
> Please add trace logging to each involved methods, include tx id, block height, if it's already included in block or not, what we are doing, and other reasonable info.
> 
> Maybe you also can add more tests.
> 
> Then you can run to verify it and check logs using testnet.
> 
> **It will help us to understand the process better and keep concensus.**

@yoxira 

Please describe.

yoxira:
> Are speaking about package.json? Doesn't it break something?

Right. Old nodes are not compatible with such version names but I think we can start using it after `spaceship` activation, as every node should be updated by this time

yoxira:
> you can run to verify it and check logs using testnet.

Does it make sense to have activation heights for consensus in a separate config file? E.g., for mainnet and testnet

adamant-al:
> > you can run to verify it and check logs using testnet.
> 
> Does it make sense to have activation heights for consensus in a separate config file? E.g., for mainnet and testnet

Hmm… Good point.
To test activation, we better move activation heights outside code, may be config is the best place.

Or may be add in a config as override (mainnet config will not have it, while testnet will).

@yoxira

yoxira:
@adamant-al 

Receiving a new tx from a client app via API (it's the first node that sees the tx):  
1. `Logic->Transaction.prototype.publish` creates a new transaction from the received data, validates the client's transaction properties (including `timestampMs`), and adds a transaction ID.
2. Processes the unconfirmed transaction as usual.

Receiving a new tx (current height) from another peer:  
1. Validates the transaction. If it's invalid, removes the peer.  
2. Processes the unconfirmed transaction as usual.

Fetching a tx from another peer when arising/verifying the blockchain from height 0:  
1. Validates the transaction. If it's invalid, removes the peer.  
2. Processes the unconfirmed transaction as usual but **verifies multiple transactions at a time**.

<h1></h1>

Processing an unconfirmed transaction flow:

1. First, `logic.transaction.process()` is called. It verifies the transaction ID and normalizes the sender ID.  
2. Then, the transaction is normalized. The `logic.transaction.objectNormalize()` method validates the transaction object against a schema and removes unnecessary properties, **including `timestampMs` before spaceship activation**.  
3. Finally, `logic.transaction.verify()` is called, verifying all properties such as `timestamp`, `timestampMs`, `signature`, etc.

adamant-al:
@yoxira

Thanks for the explanation.

I still have some questions.

> Validates the transaction…

What's the exact process? What functions does a node call?

> …If it's invalid, removes the peer.

Is the peer banned forever? For some time? Until node restart?

- Why the "Receiving a new tx from a client app via API (it's the first node that sees the tx)" doesn't have "Validates the transaction" step?

- When a node broadcast a tx to other nodes? After the "Processes the unconfirmed transaction as usual" step?

- Besides `objectNormalize()`, we have `normalize()`. Where is it used and what's the difference? Don't we add `timestampMs` there? Please add JSDoc for `normalize()`.

- Why do we have both `apply()` and `applyUnconfirmed()`? (same for `undo()`). What's the difference?

- Can you confirm there is no other tx entrance besides those three you've described?

yoxira:
@adamant-al 

> Validates the transaction... What's the exact process? What functions does a node call?

A node only calls `objectNormalize()` for schema validation of a received transaction.

> Is the peer banned forever? For some time? Until node restart?

I don't think nodes can ban peers (`BANNED` status is never applied). It just removes peer from the peer list until it rediscovers the peer again

> Why the "Receiving a new tx from a client app via API (it's the first node that sees the tx)" doesn't have "Validates the transaction" step?

Because the step is included in "processing unconfirmed transactions" flow. 

> When a node broadcast a tx to other nodes? After the "Processes the unconfirmed transaction as usual" step?

Right. 

> Besides `objectNormalize()`, we have `normalize()`. Where is it used and what's the difference? Don't we add `timestampMs` there? Please add JSDoc for normalize().

The `normalize()` method is only used for `POST /transactions/normalize` endpoint that should be deprecated.

> Why do we have both apply() and applyUnconfirmed()? (same for undo()). What's the difference?

"apply" methods change balances of the sender and recipient accounts. `applyUnconfirmed()` changes _unconfirmed balance_. Same goes for "undo" methods.

yoxira:
> Can you confirm there is no other tx entrance besides those three you've described?

When a node generates a new block, the transactions are taken from an already verified list, so everything must be OK

adamant-al:
@yoxira @bludnic 

So the "Receiving a new tx from a client app via API (1)" flow has one more pre-step "publish" compared to "Receiving a new tx from peers (2, 3)".

All of flows validates, normalizes, verifies new transactions.

**We better determine our approach for introducing blockchain changes:**

What do we have:

- Two points of Tx entrance
- The first (1) is when tx comes from a client app via API, includes `publish()`
- The second (2) is when a node gets a tx from another node. **This tx is supposed to be verified at least once.**
- Both ways run `process()`, `objectNormalize()`, `verify()`

It means in (1 - `publish()`) we can do any checks and it will not break consensus.

There is a possibility that a malicious node will broadcast a tx where checks in `publish()` may fail. ⟶ If there are important checks, we include them also in `verify()`.

Currently we have these checks in `publish()`:

- 'Transaction timestamp is in the future'
- `Transaction timestamp is more than ${constants.maxTransactionAgeSec} seconds in the past`

These checks are not in `verify()`, right? May be the reason is that developers didn't want to break consensus…

Now we also have a new check in `verify()`: `The difference between timestamp and timestampMs is greater than ${maxTimestampMsDelta}ms`.

What's our approach?

- Run same checks in `publish()` and `verify()` (after consensus activation)
- Run different checks in `publish()` and `verify()`
- Run all checks in `verify()`
- …

Probably running same checks in both functions is more secure and optimal, but how to implement it? Moving checks to `consensus.js`? Or introduce new functions like `timestampChecks(consensusSensitive?)`?

yoxira:
@adamant-al 

There is no point in adding the same checks in the `publish` method as in `verify` because the latter MUST be called after `publish`. Also, we can't include `publish`'s checks in the `verify` method because they depend on the current time. It's impossible to verify whether an old transaction is no more than 5 seconds old.

yoxira:
Regarding debug logs, I suggest making them in a separate PR. I think it's a good idea to use the debug library or a similar approach, allowing us to filter debug logs, e.g., using `DEBUG=loader npm run start`

adamant-al:
> @adamant-al
> 
> There is no point in adding the same checks in the `publish` method as in `verify` because the latter MUST be called after `publish`. Also, we can't include `publish`'s checks in the `verify` method because they depend on the current time. It's impossible to verify whether an old transaction is no more than 5 seconds old.

Will a client app receive an error text `The difference between timestamp and timestampMs is greater than ${maxTimestampMsDelta}ms.` if it's not in `publish()` but in `verify()`?

yoxira:
> Will a client app receive an error text `The difference between timestamp and timestampMs is greater than ${maxTimestampMsDelta}ms.` if it's not in `publish()` but in `verify()`?

@adamant-al Yes, as it does with `Invalid transaction timestamp. Timestamp is not in the int32 range` right now.

adamant-al (approved):
gw, lfg

adamant-al:
Tracking issue: #209

Issue URL: https://github.com/Adamant-im/adamant/issues/209

metalisk:
Follow-up: the global `maxTransactionAgeSec` admission rule introduced here is being narrowed to messaging types only in #235 (implementation: #236).

**Offline wallet note:** even after that change, types 8 and 9 still enforce the 5-second past window in `publish()`, so a future offline wallet remains blocked for chat messages and KVS updates unless clients re-sign before submission or admission policy is extended further. Non-messaging types are the intended beneficiaries of the scope change.

PR #96, opened 2025-01-04
`?returnUnconfirmed=1` flag for transactions

### LINKED ISSUE #172: [Composite] Align returnUnconfirmed across node, schema, and docs
### Summary

Align `returnUnconfirmed` behavior across ADAMANT Node, ADAMANT Schema, and public API documentation.

### Details

`returnUnconfirmed` is supported for these node endpoints:

- `/api/transactions`
- `/api/transactions/get`
- `/api/chats/get`
- `/api/chatrooms/:address`
- `/api/states/get`

This capability spans multiple companion artifacts and API expectations that should be tracked explicitly:

- `returnUnconfirmed` should be reflected in companion artifacts such as ADAMANT Schema and public API docs.
- The behavior of `returnUnconfirmed` together with `fromHeight`, `toHeight`, and `blockId` should be documented explicitly for client developers.
- We should decide whether unconfirmed transactions returned via REST should expose explicit `blockId: null`, `height: null`, and `confirmations: 0`, or whether the current omitted-field behavior should remain documented as-is.
- We should evaluate whether `/api/transactions/unconfirmed` can reuse the newer `getUnconfirmedTransactions()` query logic and eventually deprecate the older limited `getPooledTransactions()` path.

### Checklist

- [x] Create or link a companion task/PR in `Adamant-im/adamant-schema` for `returnUnconfirmed` on the affected endpoints
- [x] Update public docs/wiki/docs repo pages to describe `returnUnconfirmed` and the affected endpoints
- [x] Document filter compatibility rules for `blockId`, `fromHeight`, and `toHeight` when `returnUnconfirmed=1`
- [x] Review client compatibility for polling and pagination flows that already rely on `fromHeight`
- [x] Decide and document the unconfirmed REST response shape for `blockId`, `height`, and `confirmations`
- [x] Evaluate reusing `getUnconfirmedTransactions()` in `/api/transactions/unconfirmed` and deprecating `getPooledTransactions()`
- [x] Link all related follow-up PRs and issues back to this task

### Notes

- Related context: https://github.com/Adamant-im/adamant/pull/96
- Review thread requesting a follow-up task for unconfirmed query logic: https://github.com/Adamant-im/adamant/pull/96#discussion_r1976484742
- PR comment asking for a task to add `returnUnconfirmed` to ADAMANT Schema: https://github.com/Adamant-im/adamant/pull/96#issuecomment-2692397157
- PR comment about `fromHeight` being incompatible with `returnUnconfirmed` in existing PWA polling flow: https://github.com/Adamant-im/adamant/pull/96#issuecomment-2692532596

### Verification

- [x] Verify schema/docs examples match actual node responses for every affected endpoint
- [x] Verify at least one client flow that polls with `fromHeight` against the documented `returnUnconfirmed` behavior
- [x] Confirm there is no remaining undocumented API drift after the follow-up work is merged

yoxira:
[:shipit:  Trello task](https://trello.com/c/vrje40Ch/69-feat-return-unconfirmed-transactions-y-api-both-rest-and-socket)

Endpoints:
- [x] `/chats/get`
- [x] `/states/get`
- [x] `/chatrooms`
- [x] `/transactions`

Filters:
> [!Note]
> As unconfirmed transactions have not yet been included in the blockchain, no unconfirmed transactions will be returned when `blockId`, `fromHeight` or `toHeight` is used.

- [ ] ~~blockId~~ (_Not present in unconfirmed transactions_)
- [x] ~~fromHeight~~ (_Included in unconfirmed transactions_)
- [ ] ~~toHeight~~ (_Not present in unconfirmed transactions_)
- [x] minAmount
- [x] maxAmount
- [x] senderId
- [x] senderIds
- [x] senderPublicKey
- [x] senderPublicKeys
- [x] recipientId
- [x] recipientIds
- [x] recipientPublicKey
- [x] recipientPublicKeys
- [x] inId
- [x] isIn
- [x] type
- [x] types
- [x] key
- [x] keyIds

Options:
- [x] limit
- [x] offset
- [x] orderBy
- [x] returnAsset
- [x] withoutDirectTransfers ⟶ includeDirectTransfers


modules/chats.js
```diff
@@ -172,18 +173,52 @@ __private.list = function (filter, cb) {
   library.db.query(sql.countList({
     where: where
   }), params).then(function (rows) {
-    var count = rows.length ? rows[0].count : 0;
+    let count = rows.length ? Number(rows[0].count) : 0;
```
bludnic, line 176: Is `rows[0].count` a type string? Do you know why?

modules/transactions.js
```diff
@@ -389,6 +431,183 @@ Transactions.prototype.getUnconfirmedTransaction = function (id) {
...
+ *
+ * @param {Object} filter - Criteria to filter unconfirmed transactions
+ * @param {Object} [options]
+ * @param {string[]} [options.allowedFilters=[]] - List of keys allowed for filtering transactions
+ * @param {Object} [options.aliases={}] - Key-value pairs for aliasing filter keys
+ * @param {string} [options.defaultCondition='AND'] - Default logical condition for combining filters ('AND' or 'OR')
+ * @returns {Object[]} - Array of unconfirmed transactions matching the criteria
+ * @throws {Error} - If `filter` is not an object
+ */
+Transactions.prototype.getUnconfirmedTransactions = function (filter, options = {}) {
```
bludnic, line 524: We have `GET /api/transactions/unconfirmed` endpoint that uses `__private.getPooledTransactions()` method. Can we reuse that logic?

https://github.com/Adamant-im/adamant/blob/87b59df371c26d0a9c6fea8573bf3873ee50e18a/modules/transactions.js#L970-L972

modules/transactions.js
```diff
@@ -214,12 +216,52 @@ __private.list = function (filter, cb) {
...
       sortMethod: orderBy.sortMethod
     }), params).then(function (rows) {
-      var transactions = [];
+      let transactions = [];
 
       for (var i = 0; i < rows.length; i++) {
         transactions.push(library.logic.transaction.dbRead(rows[i]));
       }
 
+      if (filter.returnUnconfirmed) {
```
bludnic, line 225: Can we return unconfirmed transaction for `GET /api/transactions/get?id=${transactionId}` as well?

image

test/common/stubs/transactions.js
```diff
@@ -1,3 +1,196 @@
...
+    },
+    recipientId: "U5885317311990438076",
+    timestamp: 58880317,
+    signature:
+      "5ee972df476703492a667616eef428ed127e13fe5de8ba873b6579a806ddbd9fbd34147cf0321823d72e0d234466fc3dc89ebe7341e0b4a91a56b32d3bdb6a00",
+    id: "2521078418148431420",
+    fee: 50000000,
+    relays: 1,
+    receivedAt: "2019-07-16T04:38:38.492Z",
+  },
```
bludnic, line 23: Let's use `blockId: null` if its an unconfirmed transaction? What do you think? @yoxira

modules/chats.js
yoxira: because in postgresql `COUNT` returns bigint. We should probably convert it to number on SQL level, but I will do it in a separate PR for every endpoint. For now, I think I should convert the count back to string to avoid breaking changes

modules/transactions.js
yoxira: The `getPooledTransactions` method only filters by `senderPublicKey` or `address`. Do you suggest to use `getUnconfirmedTransactions` instead of `getPooledTransactions` to add complete transaction query language for those endpoints?

bludnic: Yes. For now, let's keep as is, to not break something. Later we can deprecate `getPooledTransactions` and use your solution.

test/common/stubs/transactions.js
bludnic: I mean to explicitly set `blockId` to `null` when the transaction is unconfirmed and requested via REST. Currently, it's `undefined`, so the property is omitted when converted to JSON.

modules/chatrooms.js
```diff
@@ -123,25 +123,41 @@ __private.listChats = function (filter, cb) {
...
+          isIn: filter.userId,
+        });
+
+        transactions = modules.transactions.mergeUnconfirmedTransactions(
+          transactions,
+          unconfirmedTransactions,
+          {
+            orderBy,
+            limit: params.limit,
+            offset: params.offset,
```
bludnic: Will it work properly when using `/api/chatrooms/address` with `limit=25&offset=0` params?

modules/chatrooms.js
```diff
@@ -112,11 +112,26 @@ __private.listChats = function (filter, cb) {
...
+  if (filter.returnUnconfirmed) {
+    unconfirmedTransactions = modules.transactions.getUnconfirmedTransactions({
+      isIn: filter.userId,
+    });
+
+    params.mergingOffset = unconfirmedTransactions.length;
+    params.mergingLimit = filter.limit;
+
+    params.offset = Math.max(0, params.offset - unconfirmedTransactions.length);
+    params.limit += unconfirmedTransactions.length * 2;
```
bludnic, line 127: Will it work properly when using `GET /api/chatrooms/address` with `limit=25&offset=0` params?

test/common/stubs/transactions.js
yoxira: We can add those properties in a separate PR for the API. Currently, we don't return `null` for `blockId` and `height`, so I don't see any point in changing the test stubs.

modules/chatrooms.js
yoxira: It will work, but I just realized that it fetches more transactions than needed (edit: from database)

modules/chatrooms.js
```diff
@@ -112,37 +112,64 @@ __private.listChats = function (filter, cb) {
...
+
+      if (filter.returnUnconfirmed) {
+        transactions = modules.transactions.mergeUnconfirmedTransactions(
+          transactions,
+          unconfirmedTransactions,
+          {
+            orderBy,
+            limit: params.mergingLimit,
+            offset: params.mergingOffset,
+            withoutDirectTransfers: filter.withoutDirectTransfers
```
adamantmm, line 151: `withoutDirectTransfers` is deprecated. Let's switch to `includeDirectTransfers`.

adamantmm (changes requested):

test/common/stubs/transactions.js
adamantmm: I generally like the idea to include `blockId: null` to make it clear that a tx is not confirmed.
Additionally, set `height: null` and `confirmations: 0`.
@yoxira @bludnic If you think it's reasonable and will not break anything on clients, js-api lib and console, etc, please create a Trello task for this.

adamantmm:
The client-node socket subscription currently receives unconfirmed txs, right?

yoxira:
@adamantmm yes

yoxira:
@adamant-al we already can subscribe to specific transaction types and asset types by emitting `types` and `assetChatTypes` events

test/common/stubs/transactions.js
adamant-al: @yoxira @bludnic

modules/chatrooms.js
adamant-al: Also will it influence client apps when paginating?
@yoxira

modules/chatrooms.js
adamant-al: @yoxira
What do you think?

adamant-al:
> @adamant-al we already can subscribe to specific transaction types and asset types by emitting `types` and `assetChatTypes` events

Please point to the documentation.

test/common/stubs/transactions.js
yoxira: https://trello.com/c/nuy2glkJ

modules/chatrooms.js
yoxira: @adamant-al it's not a breaking change. Requests without `returnUnconfirmed` should work as expected.

modules/chatrooms.js
yoxira: We can change this now, as we've merged the `includeDirectTransfers` PR

yoxira:
> > @adamant-al we already can subscribe to specific transaction types and asset types by emitting `types` and `assetChatTypes` events
> 
> Please point to the documentation.

https://github.com/Adamant-im/adamant/wiki/REST-and-Socket-connections#listen-for-updates

> You can also filter the transaction by the address and transaction types:
>
> ```js
> // subscribe to "transfer" and "chat message" transactions 
> socket.emit('types', [0, 8]);
> ```
> ...

Added note to improve the page in a trello task

yoxira:
Note: sorting by height doesn't work because `undefined` always return `false` in comparison. It'll work after adding `height: null` to the unconfirmed transactions. Workaround is to use `timestamp` for sorting.

adamant-al (approved):
Good work!

adamantmm (approved):

modules/transactions.js
bludnic: Please create a task for this @yoxira

bludnic:
Do we have a task for adding `returnUnconfirmed` to ADAMANT Schema? If not, could you create? @yoxira

yoxira:
@bludnic it's within the PR's task checklist

bludnic:
> As unconfirmed transactions have not yet been included in the blockchain, no unconfirmed transactions will be returned when blockId, fromHeight or toHeight is used.

We have a mechanism in PWA for retrieving new messages by polling `GET /api/chats/get`. Since we already use `fromHeight`, it is incompatible with `returnUnconfirmed`, so we cannot benefit from this feature in this particular case.

Any thoughts? @yoxira @adamant-al

```http
GET /api/chats/get/?isIn=U17430386380322813702&orderBy=timestamp:asc&fromHeight=42211162 HTTP/1.1
Host: phecda.adm.im
```

adamantmm:
> > As unconfirmed transactions have not yet been included in the blockchain, no unconfirmed transactions will be returned when blockId, fromHeight or toHeight is used.
> 
> We have a mechanism in PWA for retrieving new messages by polling `GET /api/chats/get`. Since we already use `fromHeight`, it is incompatible with `returnUnconfirmed`, so we cannot benefit from this feature in this particular case.
> 
> Any thoughts? @yoxira @adamant-al
> 
> ```httpspec
> GET /api/chats/get/?isIn=U17430386380322813702&orderBy=timestamp:asc&fromHeight=42211162 HTTP/1.1
> Host: phecda.adm.im
> ```

I offer include all of unconfirmed txs when both `fromHeight` and `returnUnconfirmed` params are set.
Apps will process unconfirmed in their own way.

`blockId`, `toHeight` is used with `returnUnconfirmed` ⟶ no unconfirmed txs
`fromHeight` is used with `returnUnconfirmed` ⟶ return unconfirmed txs

It makes sense and not a breaking change as `returnUnconfirmed=0` by default.

@yoxira @bludnic

adamant-al (approved):

PR #99, opened 2025-02-03
feat: add socket connection between nodes

### LINKED ISSUE #171: [Feat] Add socket connections between nodes
### Summary

Add socket connections between nodes so ADAMANT nodes can exchange transactions, blocks, and signatures over WebSocket in addition to the existing HTTP-based transport.

### Motivation

The current node-to-node flow relies primarily on HTTP broadcasting, which adds delay and makes peer communication less direct for real-time propagation.

Socket connections between nodes improve transaction and block delivery latency, make peer communication more responsive, and reduce reliance on one transport path while preserving backward compatibility with nodes that still use HTTP only.

### Detailed description

This feature introduces WebSocket-based communication between nodes and related peer metadata/API updates.

Expected scope:

- add WebSocket connections between nodes for broadcasting and receiving network events
- keep peer selection randomized to avoid centralization around a fixed subset of nodes
- rotate a portion of socket-connected peers periodically
- preserve compatibility with nodes that do not support the socket transport
- expose peer transport state through peer API fields such as `isBroadcastingViaSocket` and `syncProtocol`
- support fetching a specific peer via `GET /api/peers/get`
- document new config options for `wsNode`, including connection limits for broadcast and receive directions

Related merged PRs:

- [Adamant-im/adamant#99](https://github.com/Adamant-im/adamant/pull/99)
- [Adamant-im/adamant-schema#21](https://github.com/Adamant-im/adamant-schema/pull/21)

### Alternatives

- Keep HTTP-only transport and accept slower propagation of unconfirmed transactions
- Use the previous unused socket server path without implementing full receive-side node logic
- Introduce a different peer transport mechanism, which would add more complexity without clear benefit over WebSocket for the current node architecture

- introduce node-to-node WebSocket client/server handling in the node repository
- authenticate peers using node nonce and existing peer data
- track peer transport mode in the peer model and peer API responses
- add configuration for enabling socket transport and limiting broadcast/receive connection counts
- align schema and public docs with the implemented node behavior

yoxira:
https://trello.com/c/5Jt28f7V/63-feat-socket-support-node-node

## Description

- The node chooses peers to connect to in the same way as for HTTP broadcasting—randomly, to avoid centralization.
- Every 30 minutes, the node attempts to replace 20% of its WebSocket connections.
- Received transactions affect the peer's health check.
- Nodes authenticate own ip by providing nonce.

## Config

Added new required config properties:

```json5
{
  "peers": {
    // ....
    "options": {
      "maxWsConnections": 15 // maximum number of peers to receive transactions from
      // ...
    }
  },
  "wsNode": {
    "enabled": true, // enable broadcasting to other nodes using ws
    "maxConnections": 25 // maximum number of peers to broadcast transactions to
  }
}
```

## API

The `GET /api/peers` now returns info about connection type for incoming (syncing) and outcoming (broadcasting) transactions:

```js
{
  "peers": [
    {
      "isBroadcastingViaSocket": false,
      "syncProtocol": "ws" // or 'http'
    },
  ]
}
```


api/ws/server.js
```diff
@@ -0,0 +1,65 @@
...
+  this.io = new Server(server, {
+    allowEIO3: true,
+    cors: appConfig.cors,
+  });
+
+  this.enabled = appConfig.wsNode.enabled;
+  this.max = appConfig.wsNode.maxConnections;
+}
+
+WebSocketServer.prototype.linkPeers = function (logic) {
```
bludnic, line 14: We usually recommend to use Node v18 when running ADAMANT node. Can we start using classes instead of `prototype`, or could it cause issues? @yoxira @adamant-al

config.default.json
```diff
@@ -77,6 +77,7 @@
       "blackList": []
     },
     "options": {
+      "maxWsConnections": 15,
```
bludnic, line 80: Isn't better to move it to `wsNode` config and use more descriptive naming, like:

```json5
"wsNode": {
  "enabled": true,
  "maxIncomingConnections": 15, // maximum number of peers to receive transactions from
  "maxOutgoingConnections": 25, // maximum number of peers to broadcast transactions to
}
```

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
+'use strict';
+
+const { io } = require('socket.io-client');
+const Peer = require('../../logic/peer.js');
+
+const maxReconnectDelay = 60000;
+const defaultReconnectionDelay = 5000;
+
+function TransportWsApi(modules, library, options) {
```
bludnic, line 9: Can you add a JSDoc to describe the purpose of `TransportWsApi`?

app.js
```diff
@@ -606,6 +604,21 @@ d.run(function () {
       });
     }],
 
+    /**
+     * Listens for new transacttions using websocket and links peers to the websocket server
```
bludnic, line 608: typo: transacttions -> transactions

logic/peers.js
```diff
@@ -29,6 +29,18 @@ function Peers (logger, cb) {
...
 }
 
+/**
+ * Finds the peer by nonce
+ * @param {string} nonce
+ */
+Peers.prototype.getByNonce = function (nonce) {
+  for (const [peerString, peer] of Object.entries(__private.peers)) {
+    if (peer.nonce === nonce) {
+      return __private.peers[peerString];
```
bludnic, line 39:
```suggestion
      return peer;
```

Same behaviour?

bludnic: If so, you can use `Object.values` instead of `Object.entries`

config.default.json
adamant-al: It's really confusing to have `options.maxWsConnections` and `wsNode.maxConnections`.

adamant-al:
> What is the limit [of number of peers for broadcasting] for REST connections?
> It's hard coded in constants.js file:
> maxPeers: 100

Why then we have `peers.options.limits.max` in the default config?

Can you please explain what are `peers.options`, `broadcasts` and `transactions` params from config?

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+TransportWsApi.prototype.replacePeer = function(peer) {
+  const self = this;
+
+  // Remove the disconnected peer
+  self.connections.delete(`ws://${peer.ip}:${peer.port}`);
+
+  // Find a new peer to replace it
+  self.getRandomPeer((err, newPeer) => {
+    if (err || !newPeer) {
+      self.logger.debug('WebSocket: Failed to find replacement peer');
```
adamant-al, line 107: 'WebSocket: Failed to find replacement' ⟶ Add 'for peer ws://${peer.ip}:${peer.port} + err || no peers'?

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+
+  this.connections = new Map();
+
+  this.peers.events.on('peers:update', () => this.updatePeers());
+
+  this.startRotation();
+}
+
+TransportWsApi.prototype.initialize = function() {
+  const self = this;
```
adamant-al, line 28: Preferred to log this event / debug

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+
+  socket.on('blocks/change', (data) => {
+    self.transportModule.internal.postBlock(data, peer, 'websocket /blocks', (err) => {
+      if (err) {
+        self.peers.recordRequest(peer.ip, peer.port, err);
+      }
+    });
+  });
+
+  socket.on('signature/change', (data) => {
```
adamant-al, line 169: No other events possible besides
'transactions/change'
'blocks/change'
'signature/change'
?

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+    }, (err) => {
+      if (err) {
+        self.peers.recordRequest(peer.ip, peer.port, err);
+      }
+    });
+  });
+};
+
+TransportWsApi.prototype.updatePeers = function() {
+  const self = this;
```
adamant-al, line 181: Preferred to log this event / debug

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+
+  if (connection) {
+    connection.socket.removeAllListeners();
+    connection.socket.disconnect();
+    this.connections.delete(peerUrl);
+  }
+};
+
+TransportWsApi.prototype.rotatePeers = function () {
+  const self = this;
```
adamant-al, line 218: Preferred to log this event / debug

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+  const connectionsArray = Array.from(self.connections.values());
+
+  const shuffled = connectionsArray.sort(() => Math.random() - 0.5);
+
+  self.getRandomPeers(countToRotate, (err, newPeers) => {
+    if (err || !newPeers.length) {
+      return;
+    }
+
+    self.logger.debug(`Rotating ${newPeers.length} out of ${totalConnections} peers.`);
```
adamant-al, line 236: Consider to clarify logging everywhere that it's about ws.

E.g., `[wsNode] Rotating ${newPeers.length} out of ${totalConnections} peers.`

api/ws/transport.js
```diff
@@ -0,0 +1,264 @@
...
+    newPeers.forEach(newPeer => {
+      self.connectToPeer(newPeer);
+    });
+  });
+};
+
+TransportWsApi.prototype.startRotation = function() {
+  this.rotationInterval = setInterval(() => {
+    this.rotatePeers();
+  }, 1000 * 60 * 30); // 30 minutes
```
adamant-al, line 254: Make const

adamant-al:
> > The GET `/api/peers` now returns info about connection type for incoming (syncing) and outcoming (broadcasting) transactions:
> 
> ```
> {
>   "peers": [
>     {
>       "isBroadcastingViaSocket": false,
>       "syncProtocol": "ws" // or 'http'
>     },
>   ]
> }
> ```
> 
> If `syncProtocol` is "ws", it means the node accepts new txs both via ws and http? Maybe then something like `"syncProtocol": "wsAndHttp" // or 'http'`?

I see that current implementation assumes that a node is connected either via ws, or http.

One of reasons while adding wsNode is redundancy. It means some nodes may connect with ws AND http. Imaging that a network has low amount of nodes, and all of them ws-enabled.

Can we improve that some nodes will connect both ws and http?

modules/peers.js
```diff
@@ -299,6 +303,28 @@ Peers.prototype.update = function (peer) {
   return library.logic.peers.upsert(peer);
 };
 
+/**
+ * Changes the connection type to ws
+ * @param {Peer} peer
+ */
+Peers.prototype.switchToWs = function (peer) {
```
adamant-al, line 310: See my comment about redundancy
https://github.com/Adamant-im/adamant/pull/99#issuecomment-2672255247

api/ws/transport.js
adamant-al: May be 'rounds/change'?

adamant-al (changes requested):

api/ws/transport.js
yoxira: `rounds/change` is not used.

yoxira:
> Does REST node exchange uses the same authentication? What's it for?

The authentication is needed to link the incoming WebSocket connection to the existing peer. REST provides `port` instead.

> Does the node get its nonce once and forever, updates it on restart, updates it on some event?

A nonce is generated on app start and never changes. 

> Is the node's nonce stays the same across all nodes?

Yes


> Please elaborate:

>> Received transactions affect the peer's health check.

If a node receives an invalid transaction over WebSocket, the peer will be removed from the peer list, as it would be using REST.

yoxira:
> I see that current implementation assumes that a node is connected either via ws, or http.

A node can be connected using both `ws` and `http` at the same time. It's up to a broadcasting node to decide whether to double-send the transactions or not. In the current implementation, a node doesn't send the transactions twice at all, while the old nodes always do.

I suggest still preferring nodes with http-only connections when broadcasting, but also choosing some % of ws-connected nodes to broadcast to using http.

adamant-al:
@yoxira 

> A node can be connected using both `ws` and `http` at the same time. It's up to a broadcasting node to decide whether to double-send the transactions or not. In the current implementation, a node doesn't send the transactions twice at all, while the old nodes always do.

Please explain how an old node and new node broadcast txs (both new received from a client app and not new, received from other node).

> I suggest still preferring nodes with http-only connections when broadcasting, but also choosing some % of ws-connected nodes to broadcast to using http.

This is not clear as well.

adamant-al:
> > What is the limit [of number of peers for broadcasting] for REST connections?
> > It's hard coded in constants.js file:
> > maxPeers: 100
> 
> Why then we have `peers.options.limits.max` in the default config?
> 
> Can you please explain what are `peers.options`, `broadcasts` and `transactions` params from config?

Please explain @yoxira

yoxira:
@adamant-al 

After verifying an unconfirmed transaction (whether it comes from another node or a client), a node adds it to the queue for the next broadcast (broadcastInterval is set in the config file). When the node releases the queue, it sends POST requests to random peers with the unconfirmed (but verified and unmodified) transaction

Before this PR, the node sent transactions randomly to any connected peers without filtering. This meant the nodes could send transactions via HTTP to a peer connected via WebSocket, making the PR version process the same transaction from both HTTP and WebSocket

In this PR, I made a small change to prefer HTTP connections when broadcasting, as they rely only on HTTP to receive transactions. Each broadcast release still selects WebSocket peers, but HTTP-connected nodes now have a slightly higher chance of being chosen

api/ws/server.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+class WebSocketServer {
+  constructor(server, appConfig) {
+    this.io = new Server(server, {
+      allowEIO3: true,
+      cors: appConfig.cors,
+    });
+
+    this.enabled = appConfig.wsNode.enabled;
+    this.max = appConfig.wsNode.maxIncomingConnections;
```
adamant-al: Is it correct that it's `maxIncomingConnections`?
It's "server to broadcast"

api/ws/transport.js
```diff
@@ -0,0 +1,350 @@
...
+class TransportWsApi {
+  constructor(modules, library, options) {
+    this.modules = modules;
+    this.library = library;
+    this.peers = modules.peers;
+    this.system = modules.system;
+    this.transportModule = modules.transport;
+    this.logger = library.logger;
+
+    this.maxConnections = options.maxOutgoingConnections;
```
adamant-al: Is it correct that it's `maxOutgoingConnections`?
It's "WebSocket to receive"

yoxira:
@adamant-al 

- `peers.options` – Request limiter parameters. These do not limit outgoing transactions but restrict incoming requests
- `broadcasts`  
  - `broadcastInterval` – The interval (in ms) at which the node broadcasts unconfirmed transactions
  - `broadcastLimit` – The number of peers a node broadcasts to per interval. This overrides the `maxPeers` constant from the config file
  - `parallelLimit` – The maximum number of asynchronous requests the node can make when broadcasting
    - Example: If `parallelLimit = 5` and `broadcastLimit = 10`, the node will first send transactions to 5 peers asynchronously, then process another 5
  - `releaseLimit` – The number of transactions (or other data) included in each broadcast
    - Example: If 10 transactions are waiting in the queue and `releaseLimit = 5`, only 5 will be sent in the next broadcast
  - `relayLimit` – The maximum number of nodes that can process an unconfirmed transaction
    - Example: If a transaction has already been broadcasted to 5 nodes and `relayLimit = 4`, the node will process the transaction but won't broadcast it further
- `transactions.maxTxsPerQueue` – The maximum number of unprocessed transactions allowed in the queue. If the number of pending transactions exceeds `maxTxsPerQueue`, the node will ignore new incoming transactions

adamant-al:
@yoxira 

Why don't we set default values for `peers.options.limits` and `api.options.limits` to protect nodes from abuse?

api/ws/server.js
yoxira: I was thinking about the parameter names earlier, and it's still confusing to me because `maxIncomingConnections` sounds like it refers to connections coming into the server, while `maxOutgoingConnections` sounds like connections going out from the server

yoxira: `maxToReceiveFrom` and `maxToBroadcastTo` are more clear but ugly alternatives

adamant-al:
@yoxira 

Please confirm:

- A node broadcasts a single unconfirmed transaction to several nodes
- A node broadcasts a single unconfirmed transaction both via rest and via socket
- A node broadcasts a single unconfirmed transaction to a single node via rest or via socket, but never both ways
- A single node may receive a single unconfirmed transaction several times from different nodes (redundancy)
- A single node may receive a single unconfirmed transaction from several nodes (redundancy)
- A single node may receive a single unconfirmed transaction both via rest and via socket (redundancy)
- Txs broadcasting via rest have a delay of `broadcastInterval`
- Txs broadcasting via socket is instant with no delays
- Before this PR broadcasting via socket also happened, but in a different way (what changed?)
- After this PR we ensure that a single unconfirmed transaction spreads over ADM network instantly even if most nodes use previous software

yoxira:
@adamant-al 

It was like this for more than 7 years according to git history. The options are in legacy mode and should probably be updated to match last version of `express-rate-limiter`.

yoxira:
> A node broadcasts a single unconfirmed transaction to several nodes

Yes, randomly chosen ones

> A node broadcasts a single unconfirmed transaction both via rest and via socket

Yes, depending on configuration

> A node broadcasts a single unconfirmed transaction to a single node via rest or via socket, but never both ways

No. If a peer is connected via socket, it may receive a transaction via rest  

> A single node may receive a single unconfirmed transaction several times from different nodes (redundancy)

Yes

> A single node may receive a single unconfirmed transaction from several nodes (redundancy)

Yes

> A single node may receive a single unconfirmed transaction both via rest and via socket (redundancy)

Yes.

> Txs broadcasting via rest have a delay of `broadcastInterval`

Yes. 

> Txs broadcasting via socket is instant with no delays

Yes. 

> Before this PR broadcasting via socket also happened, but in a different way (what changed?)

The old versions had a separate WebSocket server for nodes, but it wasn't used: there was no logic to receive the transactions via socket

> After this PR we ensure that a single unconfirmed transaction spreads over ADM network instantly even if most nodes use previous software

The new nodes will receive transactions instantly regardless of the other node versions. The old ones will still rely on broadcasting when receiving transactions. See the answer above.

adamant-al:
@yoxira 

>> A node broadcasts a single unconfirmed transaction to a single node via rest or via socket, but never both ways

> No. If a peer is connected via socket, it may receive a transaction via rest

I mean a single node when broadcasting a single unconfirmed transaction to a single node, it will use socket OR rest, but not both.

Or it can broadcast it via socket and then again via rest (specific one tx to a single node)?

api/ws/server.js
adamant-al: maxBroadcastConnections
maxReceiveConnections
?

adamant-al:
@yoxira 

Can we add tests for this PR?

yoxira:
> Or it can broadcast it via socket and then again via rest (specific one tx to a single node)?

Yes, it can.

yoxira:
> Can we add tests for this PR?

This will require a lot of stubs and can become testing stubs instead of an actual logic really quickly

api/ws/server.js
adamant-al: @yoxira @bludnic 
What do you think?

adamant-al (approved):

PR #108, opened 2025-04-07
fix: transaction pool doesn't propertly remove transactions

### LINKED ISSUE #174: [Bug] Transaction pool does not properly remove transactions
## Summary

The transaction pool keeps per-queue transaction arrays and ID indexes that can drift out of sync after removals. This can leave stale entries in internal state, make duplicate detection unreliable for some transactions, and expose incorrect pool counters through shared transaction statistics.

## Steps to reproduce

1. Add a transaction to the unconfirmed, queued, bundled, or multisignature pool.
2. Remove that transaction and continue adding or removing entries in the same queue.
3. Query the queue contents or transaction counters, or check whether the pool already contains a transaction whose stored index is `0`.
4. Observe that removed entries remain as tombstones, duplicate detection can miss queue members, and counters may no longer reflect active transactions only.

## Expected behavior

Removing a transaction should delete it fully from the corresponding pool. Queue membership checks should work for every valid position, including the first element. Pool counters and exported statistics should reflect only active transactions without periodic queue reindexing.

## Actual behavior

Removed transactions are replaced with placeholder values instead of being deleted cleanly. Queue membership checks rely on truthy index values, so transactions stored at index `0` can be treated as absent. Queue counters can include stale entries, and consistency depends on periodic queue reindexing.

## Screenshots or videos

N/A

## Environment

- Component: ADAMANT Node transaction pool
- Affected files: `logic/transactionPool.js`, `modules/transactions.js`
- Reproducibility: independent of OS; tied to in-memory pool bookkeeping

## Related PR

- https://github.com/Adamant-im/adamant/pull/108

yoxira:
Rewrite the transaction pool to use `Map()` instead of manually update "indexes" and arrays.

### DIFF
--- a/logic/transactionPool.js
+++ b/logic/transactionPool.js
@@ -40,14 +40,14 @@ function TransactionPool (broadcastInterval, releaseLimit, transaction, bus, log
   };
   self = this;
 
-  self.unconfirmed = { transactions: [], index: {} };
-  self.bundled = { transactions: [], index: {} };
-  self.queued = { transactions: [], index: {} };
-  self.multisignature = { transactions: [], index: {} };
+  self.unconfirmed = new Map();
+  self.bundled = new Map();
+  self.queued = new Map();
+  self.multisignature = new Map();
+
   self.expiryInterval = 30000;
   self.bundledInterval = library.config.broadcasts.broadcastInterval;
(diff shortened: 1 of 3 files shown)


yoxira:
Live on [ahead.adamant.im](https://ahead.adamant.im)

adamant-al (approved):

PR #114, opened 2025-06-02
fix: `includeDirectTransfers` 

yoxira:
The default value for chatroom endpoints should be `1`.
> Source: https://docs.adamant.im/api/transactions-query-language.html#includedirecttransfers

Additionally, `includeDirectTransfers` was mistakenly working as an exclude option for the `/api/chatrooms/{address}` endpoint


adamant-al (approved):
Good finding @yoxira

confleux (approved):

PR #111, opened 2025-05-21
Update to nodejs 20, 22 & Update dependencies

graycraft:

PR #116, opened 2025-06-05
chore: update dependencies

yoxira:
- `bitcore-mnemonic` – No changelog yet. A new version was published this morning in npm, but their GitHub repository still contains the old version
- `chai`, `change-case`, `execa` – These packages have dropped support for commonjs
- `rimraf` – Downgraded to match our target Node.js versions. The latest version only includes dependency updates, so we’re not missing anything
- `nyc` – Completely removed as it's used only through `npx` command which downloads the package by itself


yoxira:
closed in favor of #111

PR #117, opened 2025-06-06
BREAKING CHANGE: drop Node.js 18

yoxira:
and update installation scripts

### DIFF
--- a/tools/install_node.sh
+++ b/tools/install_node.sh
@@ -7,7 +7,7 @@ databasename="adamant_main"
 configfile="config.json"
 processname="adamant"
 port="36666"
-nodejs="hydrogen"
+nodejs="jod"
 
 while getopts 'b:n:j:' OPTION; do
   OPTARG=$(echo "$OPTARG" | xargs)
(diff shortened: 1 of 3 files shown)


tools/install_node.sh
```diff
@@ -31,12 +31,12 @@ while getopts 'b:n:j:' OPTION; do
       fi
       ;;
     j)
-      if [ "$OPTARG" == "16" ] || [ "$OPTARG" == "gallium" ]
+      if [ "$OPTARG" == "20" ] || [ "$OPTARG" == "iron" ]
```
adamantmm, line 34: Will we support v24?

confleux:
Is that right that there are no deps, which would restrict v20 usage?

adamant-al:
@yoxira 
Update please "printf" messages,
Especially script version v2.1.3 ⟶ v2.2.0.

yoxira:
@confleux dependencies are updated in #111, everything was tested in node.js v20 and v22

confleux (approved):

tools/install_node.sh
yoxira: I suggest supporting only LTS versions (Node.js v24 is expected to become LTS in October)

adamant-al:
@yoxira 
Please update

adamant-al (approved):

PR #120, opened 2025-06-29
fix: skipping delegate slot

### LINKED ISSUE #180: [Bug] Node may forge or process blocks before initial sync readiness is established
### Summary

The node may attempt to forge a block or handle an incoming block before initial sync readiness is established. During startup this can produce repeated `Skipping delegate slot` warnings and may also surface discarded-block warnings while the loader has not yet confirmed whether sync is required.

### Steps to reproduce

1. Run an ADAMANT node with forging enabled and reachable peers.
2. Restart the node and let it enter the startup sequence where peers are ready but initial sync readiness is not yet explicitly confirmed.
3. Observe the forging and block-processing paths during that window.
4. Check the logs for repeated startup warnings such as:

```text
Skipping delegate slot
Skipping delegate slot
Skipping delegate slot
```

5. In some cases, also observe `Discarded block that does not match with current chain` warnings during the same startup race.

### Expected behavior

The node must not try to forge a block or process new incoming blocks until the loader has completed the initial sync-readiness check and the client is ready to participate in normal block flow.

### Actual behavior

The forging and block-receive paths can evaluate readiness too early by checking active syncing state without a separate readiness flag. As a result, the node can enter delegate-slot or block-handling logic before startup synchronization state is fully established.

### Screenshots or videos

N/A

### Environment

ADAMANT Node `dev` branch during node startup on a forging-enabled node with peers available.

### See also

- Related PR: https://github.com/Adamant-im/adamant/pull/120

yoxira:
Fixed a bug where the blockchain tried to forge a delegate before syncing had started leading to the warnings:

```
Skipping delegate slot
Skipping delegate slot
Skipping delegate slot
Skipping delegate slot
Skipping delegate slot
Skipping delegate slot
```

### DIFF
--- a/modules/blocks/process.js
+++ b/modules/blocks/process.js
@@ -368,7 +368,8 @@ Process.prototype.onReceiveBlock = function (block) {
   library.sequence.add(function (cb) {
     // When client is not loaded, is syncing or round is ticking
     // Do not receive new blocks as client is not ready
-    if (!__private.loaded || modules.loader.syncing() || modules.rounds.ticking()) {
+    const syncPending = !modules.loader.isReadyToSync() || modules.loader.syncing();
+    if (!__private.loaded || syncPending || modules.rounds.ticking()) {
       library.logger.debug('Client not ready to receive block', block.id);
       return;
     }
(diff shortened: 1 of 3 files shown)


yoxira:
This should also solve "discarded the received block"

adamantmm (approved):

confleux (approved):

PR #127, opened 2025-07-02
feat: return additional fields for unconfirmed transactions

### LINKED ISSUE #182: [Enhancement] Use timestampMs fallback when sorting transactions by timestamp
### Summary

Support sorting by `timestampMs` with fallback to `timestamp` when transaction lists are ordered by `timestamp`.

### Motivation

Millisecond precision should be used when it is available, while preserving compatibility for records and code paths that still rely on `timestamp`.

### Detailed description

- Apply `timestampMs`-first sorting for `orderBy=timestamp:asc|desc`.
- Use `timestamp` as the fallback sort field when `timestampMs` is missing.
- Keep the default transaction ordering consistent with the existing `timestamp DESC` behavior.
- Preserve current behavior for all non-`timestamp` sort fields.
- Keep the behavior aligned across SQL-backed transaction lists and unconfirmed transaction flows.

### Alternatives

- Continue sorting only by `timestamp` and ignore `timestampMs`.
- Rework the entire sorting pipeline for generic multi-field support instead of handling the `timestamp` case directly.

- Extend shared order-by formatting so SQL-backed sorting can emit `"timestampMs" <direction>, "timestamp" <direction>` for timestamp-based ordering.
- Keep default sort field and sort method initialization in transaction list code paths.
- Add regression coverage for confirmed and unconfirmed transaction listing scenarios that depend on timestamp ordering.

### Related pull requests

- https://github.com/Adamant-im/adamant/pull/127
- https://github.com/Adamant-im/adamant/pull/128

yoxira:
The properties are changed only in endpoint responses, to avoid breaking the logic


confleux (approved):

modules/transactions.js
```diff
@@ -184,7 +184,9 @@ __private.list = function (filter, cb) {
   }
 
   var orderBy = OrderBy(
-      filter.orderBy, {
+    filter.orderBy, {
+        sortField: 'timestamp',
```
adamant-al, line 188: Consider to use `timestampMs` with fallback to `timestamp`.
If we don't do it now, we'll forget later.

yoxira: We don't have the functionality to fallback sort fields yet, I will create a separate PR for this feature.

yoxira: https://github.com/Adamant-im/adamant/pull/128

PR #135, opened 2025-08-06
Fix/minor bugs

### LINKED ISSUE #146: [Bug]  WsNodeServer doesn't work properly (no proper logging)
### Summary

It seems that `WsNodeServer` does not work as expected

### Steps to reproduce

1. Run two ADAMANT nodes (local + remote)
2. Enable WsNode logging
3. Observe logs during peer connections

### Expected behavior

When a WebSocket peer connects, `WsNodeServer` should output a log message, e.g.:

```
[WsNodeServer] WebSocket peer ${peerIp} is connecting…
```

### Actual behavior

No such messages are present in logs. Appears that `WsNodeServer` is not functioning correctly.

### Possible Causes
- Logging not properly initialized for `WsNodeServer`
- `WsNodeServer` not actually accepting/handling connections

### Environment

- Tried running two nodes: one local and one remote
- Logs checked on both, no `WsNodeServer` messages appear

yoxira:


modules/node.js
```diff
@@ -65,12 +65,6 @@ Node.prototype.onBind = function (scope) {
   };
 };
 
-/**
- * Triggers onPeersReady after:
- * - Ping to every member of peers list.
- * - Load peers from database and checks every peer state and updated time.
- * - Discover peers by getting list and validates them.
- */
 Node.prototype.onBlockchainReady = function () {
```
adamant-al, line 68: Why removing JSDoc?

modules/node.js
```diff
@@ -135,10 +129,10 @@ Node.prototype.shared = {
             epoch: constants.epochTime,
             height: lastBlock.height,
             fee: library.logic.block.calculateFee(),
-            milestone: __private.blockReward.calcMilestone(lastBlock.height),
+            milestone: lastBlock.height ?? __private.blockReward.calcMilestone(lastBlock.height),
```
adamant-al: Is it necessary?
Node returns these values up to it's current height (even if still in sync)

modules/node.js
yoxira: The comment doesn't represent the function body. It looks like it was copy-pasted from the `modules/peers.js` method, that does trigger `onPeersReady`:

```js
Peers.prototype.onBlockchainReady = function () {
  // ...
  library.bus.message('peersReady');
}
```

modules/node.js
yoxira: if `lastBlock.height` is `undedfined` (e.g., before first block is loaded), then it will throw an error. Other endpoints return "Blockchain is still loading" error in such case, but the `/api/node/status` endpoint is used for tracking sync and blockchain rebuild, so it should return values even before blockchain is loaded.

modules/node.js
adamant-al: Write correct JSDoc instead of deleting

modules/node.js
adamant-al: So,

⟶ before first block is loaded (very short period of time)
Before: Endpoint /status throws an error
Now: Endpoint /status returns undefined for reward, supply and milestone

⟶ after first block is loaded (any height >= 1, even in sync)
Before&Now: Endpoint /status returns values up to synced height

It's important because apps use this endpoint for health check, even for in-sync nodes.

yoxira: @adamant-al Yes, that's how it works now.

modules/node.js
yoxira: The function is empty again, so there's nothing to describe. I can revert the removal for convention

adamant-al (approved):

PR #137, opened 2025-08-11
Fix SQL migrations

confleux:
## `20241218010631_addTimestampMs.sql`

It's just about idempotency

## `20250417154132_mergeBlockIntoTransactions.sql`

`ADD COLUMN` with constraint `NOT NULL` fails if table is not empty:

```text
ERROR:  column "height" of relation "trs" contains null values
```

Specifying default value resolves the issue.

Steps to reproduce locally:

```text
=# create table add_column_test(id int);
CREATE TABLE
=# insert into add_column_test values (1);
INSERT 0 1
=# alter table add_column_test add column "new" int not null;
ERROR:  column "new" of relation "add_column_test" contains null values
=# alter table add_column_test add column "new" int not null default 0;
ALTER TABLE
```

### DIFF
--- a/sql/migrations/20241218010631_addTimestampMs.sql
+++ b/sql/migrations/20241218010631_addTimestampMs.sql
@@ -5,7 +5,7 @@
 BEGIN;
 
 -- Add timestampMs column to Transactions
-ALTER TABLE "trs" ADD COLUMN "timestampMs" BIGINT;
+ALTER TABLE "trs" ADD COLUMN IF NOT EXISTS "timestampMs" BIGINT;
 
 -- Add timestampMs index
 CREATE INDEX IF NOT EXISTS "timestamp_ms" ON "trs"("timestampMs");
--- a/sql/migrations/20250417154132_mergeBlockIntoTransactions.sql
+++ b/sql/migrations/20250417154132_mergeBlockIntoTransactions.sql
@@ -5,8 +5,8 @@
 BEGIN;
 
 -- Create new columns and indexes
-ALTER TABLE "trs" ADD COLUMN "height" INT;
-ALTER TABLE "trs" ADD COLUMN "blockTimestamp" INT;
+ALTER TABLE "trs" ADD COLUMN "height" INT DEFAULT 0;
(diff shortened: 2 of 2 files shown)


yoxira:
@confleux We must ensure that height is correct everywhere instead of falling back to the default value. It looks like your database might be corrupted if it produces null values after updating the height of existing transactions.

confleux:
> @confleux We must ensure that height is correct everywhere instead of falling back to the default value. It looks like your database might be corrupted if it produces null values after updating the height of existing transactions.

Still, shouldn't `trs` table have rows before applying migration?

yoxira:
@confleux

If there are no rows in the `trs` table before the migration, this script does nothing.
If rows do exist, their values should be populated from the blocks table:

```sql
-- Update existing transactions
UPDATE trs
SET height = blocks.height,
    "blockTimestamp" = blocks.timestamp
FROM blocks
WHERE trs."blockId" = blocks.id;
-- Enforce NOT NULL after data is valid
ALTER TABLE "trs" ALTER COLUMN "height" SET NOT NULL;
ALTER TABLE "trs" ALTER COLUMN "blockTimestamp" SET NOT NULL;
```

confleux:
> @confleux
> 
> If there are no rows in the `trs` table before the migration, this script does nothing. If rows do exist, their values should be populated from the blocks table:
> 
> ```sql
> -- Update existing transactions
> UPDATE trs
> SET height = blocks.height,
>     "blockTimestamp" = blocks.timestamp
> FROM blocks
> WHERE trs."blockId" = blocks.id;
> -- Enforce NOT NULL after data is valid
> ALTER TABLE "trs" ALTER COLUMN "height" SET NOT NULL;
> ALTER TABLE "trs" ALTER COLUMN "blockTimestamp" SET NOT NULL;
> ```

Fine, just found out I've checked out at tag `v0.10.0`, which doesn't have a [commit](https://github.com/Adamant-im/adamant/commit/bf827efee6febcb911716b2908fe54f22d3314e9) fixing this and also missed, there is no `NOT NULL` at the latest `dev` state. Sorry for interrupting

PR #139, opened 2025-08-16
Update default nodes

adamant-al:
* Delete 207.244.243.23
* Add 5.104.87.219
* Add 200.69.21.176
* Delete 157.90.229.236 (testnet)


adamant-al:
@yoxira Ps review

yoxira (approved):

PR #142, opened 2025-08-25
feat: set missing config values to values from `config.default.json`

### LINKED ISSUE #191: [Enhancement] Populate missing config values from config.default.json during node startup
## Summary

Automatically populate missing keys in `config.json` from `config.default.json` during node startup while preserving explicitly configured values.

## Motivation

Node upgrades can introduce new configuration parameters. Operators who keep an older `config.json` may miss newly added keys, which increases manual migration work and can lead to startup or validation problems after an update.

## Detailed description

- Load `config.default.json` when it is available.
- Merge only missing keys from the default config into the user-provided config.
- Preserve all values explicitly defined by the node operator.
- Apply the merge recursively for nested configuration objects.
- Keep behavior safe when `config.default.json` is absent.
- Preserve strict validation for genuinely invalid configuration data.

## Alternatives

- Require operators to manually compare and update `config.json` on each release.
- Maintain separate migration logic for every newly introduced config field.
- Relax schema validation for missing fields, which would hide real misconfiguration.

- Update `helpers/config.js` to load runtime and default config files separately.
- Add a helper that recursively fills only missing keys.
- Apply the merge before schema validation.
- Keep existing validation and `validateForce(configData)` semantics unchanged.

## Related PR

- #142

yoxira:
This feature helps with migrations from previous versions that introduced new config parameters.

### DIFF
--- a/helpers/config.js
+++ b/helpers/config.js
@@ -14,24 +14,36 @@ var constants = require('../helpers/constants.js');
  * @return {Object} configData
  */
 function Config (configPath) {
-  var configData = fs.readFileSync(path.resolve(process.cwd(), (configPath || 'config.json')), 'utf8');
+  try {
+    const configJson = fs.readFileSync(path.resolve(process.cwd(), (configPath || 'config.json')), 'utf8');
+    const defaultConfigJson = fs.readFileSync(path.join(__dirname, '../config.json'), 'utf8');
 
-  if (!configData.length) {
-    console.log('Failed to read config file');
-    process.exit(1);
-  } else {
-    configData = JSON.parse(configData);
-  }
+    if (!configJson.length) {
+      throw 'Failed to read config file';
+    }
 
-  var validator = new z_schema();
-  var valid = validator.validate(configData, configSchema.config);
+    if (!defaultConfigJson.length) {
+      throw 'Default config is missing';
+    }
 
-  if (!valid) {
-    console.log('Failed to validate config data', validator.getLastErrors());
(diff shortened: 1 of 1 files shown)


helpers/config.js
```diff
@@ -14,24 +14,36 @@ var constants = require('../helpers/constants.js');
  * @return {Object} configData
  */
 function Config (configPath) {
-  var configData = fs.readFileSync(path.resolve(process.cwd(), (configPath || 'config.json')), 'utf8');
+  try {
+    const configJson = fs.readFileSync(path.resolve(process.cwd(), (configPath || 'config.json')), 'utf8');
+    const defaultConfigJson = fs.readFileSync(path.join(__dirname, '../config.json'), 'utf8');
```
adamant-al: '../config.default.json'?

helpers/config.js
```diff
@@ -14,24 +14,36 @@ var constants = require('../helpers/constants.js');
...
-  } else {
-    configData = JSON.parse(configData);
-  }
+    if (!configJson.length) {
+      throw 'Failed to read config file';
+    }
 
-  var validator = new z_schema();
-  var valid = validator.validate(configData, configSchema.config);
+    if (!defaultConfigJson.length) {
```
adamant-al: User may have no default config, it's ok.
Merge only when default config, don't throw.

adamant-al (changes requested):

adamant-al (approved):

PR #241, opened 2026-06-15
Chore: Modernize node installation and repair scripts

### LINKED ISSUE #240: [Enhancement] Modernize node installation and repair scripts
## Summary

Modernize the supported ADAMANT node installation and repair scripts for current VPS environments while making repeated runs safer and easier to understand.

## Motivation

The existing scripts are tied to older operating-system and runtime versions, assume a mostly clean host, and can overwrite configuration or fail when users, databases, repositories, or packages already exist.

Node operators need a maintained installation path that supports current Ubuntu and RHEL-compatible VPS images, uses supported Node.js releases, preserves existing installations where possible, and provides actionable logs when a step fails.

## Detailed description

Update the Linux installation and repair tooling to:

- support Ubuntu 20.04, 22.04, 24.04, and 26.04 LTS;
- support RHEL-compatible releases 8-10, including CentOS Stream, Rocky Linux, AlmaLinux, and RHEL;
- support Node.js 22, 24, and 26, with Node.js 24 as the default;
- update nvm, PostgreSQL installation sources, PM2, and related system packages;
- preserve existing system users, databases, PostgreSQL major versions, ADAMANT configuration files, and local Git changes during repeated runs;
- avoid loading a blockchain image into a non-empty database;
- validate a repair snapshot before dropping the existing database;
- improve error handling, comments, messages, persistent logging, and command-line help;
- document supported platforms, repeat-run behavior, logs, and the destructive repair workflow in `README.md`.

The changes are operational only and must not modify consensus, serialization, protocol, or blockchain validation behavior.

## Alternatives

- Keep the existing scripts and document manual per-distribution workarounds. This leaves operators with outdated package sources and inconsistent rerun behavior.
- Replace the scripts with a larger configuration-management or container deployment system. That may be useful separately, but it would be a substantially larger change and would not improve the existing lightweight VPS installation path.

- Detect and validate supported operating systems before making changes.
- Use official PostgreSQL APT/YUM repositories, including the PGDG archive required by Ubuntu 20.04.
- Install PostgreSQL 18 on new supported systems, while preserving an already installed major version; use PostgreSQL 17 from the archive on Ubuntu 20.04.
- Install nvm 0.40.5 and allow Node.js 22, 24, or 26.
- Make role, database, repository, configuration, PM2, and blockchain-image setup idempotent where practical.
- Keep logs under `/var/log` and fail clearly when a required operation does not complete.
- Validate the scripts with Bash syntax checks, ShellCheck, option parsing checks, and external source URL checks.

## Verification

- `bash -n tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh`
- `npx --yes shellcheck@4.1.0 tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh`
- `git diff --check origin/dev...HEAD`
- Verify `-h`, invalid network, unsupported Node.js, missing argument, and unknown option handling.
- Verify availability of the nvm installer, PostgreSQL repositories and key, and mainnet/testnet blockchain image URLs.
- Review installation on disposable Ubuntu and RHEL-compatible VPS environments before release.

## Related pull request

https://github.com/Adamant-im/adamant/pull/241

metalisk:
## Description

Modernize the ADAMANT node installation and repair scripts for current VPS environments.

The branch updates:

- `tools/install_node.sh` for Ubuntu 20.04, 22.04, 24.04, and 26.04 LTS;
- `tools/install_node_centos.sh` for RHEL-compatible releases 8-10;
- `tools/fix_node.sh` to validate a downloaded snapshot before replacing the database;
- `README.md` with current support, rerun behavior, logging, and repair guidance.

Both installers now support Node.js 22, 24, and 26, default to Node.js 24, install nvm 0.40.5, use current official PostgreSQL sources, and preserve existing users, databases, PostgreSQL major versions, configuration files, and local Git changes where practical.

Repeated installation skips blockchain image import when the selected database already contains tables. The repair tool remains intentionally destructive, but it now completes preflight checks and validates the downloaded gzip image before stopping the node and dropping the database.

## Related issue

Closes #240

## External links (optional)

- [Node.js release schedule](https://nodejs.org/en/about/previous-releases)
- [nvm releases](https://github.com/nvm-sh/nvm/releases)
- [Ubuntu release cycle](https://ubuntu.com/about/release-cycle)
- [PostgreSQL APT repository](https://apt.postgresql.org/)
- [PostgreSQL RPM repository](https://www.postgresql.org/download/linux/redhat/)

Not applicable.

## Breaking changes

No consensus, protocol, serialization, API, or storage-schema behavior changes are introduced.

The installer no longer offers Node.js 20 because the node now requires Node.js 22.13.0 or newer. Node.js 24 becomes the default. New PostgreSQL installations use PostgreSQL 18 where available, or PostgreSQL 17 on Ubuntu 20.04; existing PostgreSQL major versions are preserved.

The installers update system packages and configure services, so operators should review backups and custom package/service configuration before running them on an existing server.

## How to test

Executed:

```sh
bash -n tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh
npx --yes shellcheck@4.1.0 tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh
git diff --check origin/dev...HEAD
```

Also verified:

- `-h` output for all three scripts;
- invalid network, unsupported Node.js, missing argument, and unknown option handling;
- availability of nvm 0.40.5, the PostgreSQL signing key and supported APT repositories, PGDG RPM repository packages, and both blockchain snapshot URLs.

Not executed:

- full installation or repair against live VPS instances, because these workflows update system packages and services and can replace a PostgreSQL database.

Recommended QA:

1. Run each installer on a disposable clean supported VPS.
2. Run it again with an existing user, repository, configuration, PostgreSQL database, and PM2 process.
3. Confirm local Git changes and existing configuration values are preserved except for the requested database password update.
4. Confirm a non-empty database prevents blockchain image import.
5. Run `fix_node.sh` against a disposable node and confirm the snapshot is downloaded and validated before database replacement.
6. Reboot and confirm PM2 restores the configured node.

## Notes for reviewers

Please pay particular attention to:

- PostgreSQL repository selection on Ubuntu 20.04 versus newer Ubuntu releases;
- detection and preservation of existing PostgreSQL installations;
- RHEL-compatible PostgreSQL service and binary discovery;
- Git/configuration preservation during repeated runs;
- PM2 update, startup, and graceful stop/restart behavior;
- snapshot validation and the ordering of destructive repair operations;
- database password handling and avoidance of secret values in logs.

This branch does not touch consensus-critical node logic.

## Checklist

- [x] Code is formatted
- [x] Tests added/updated (not needed for shell-only operational tooling)
- [x] Documentation updated
- [ ] Changes tested in all relevant environments

### DIFF
--- a/tools/fix_node.sh
+++ b/tools/fix_node.sh
@@ -22,6 +22,24 @@ usage() {
   printf "Usage: %s [-h] [-n mainnet|testnet]\n" "${0##*/}"
 }
 
+require_interactive_tty() {
+  if [[ ! -r /dev/tty || ! -w /dev/tty ]]; then
+    printf "\nThis repair tool requires an interactive terminal for confirmation.\n" >&2
+    printf "Run it from a normal shell session, for example:\n" >&2
+    printf "    curl -fsSL https://adamant.im/fix_node.sh | sudo bash -s -- -n mainnet\n\n" >&2
+    exit 1
+  fi
+}
+
+read_from_tty() {
+  local prompt_text="$1"
+  local response
+
+  printf "%s" "$prompt_text" > /dev/tty
+  IFS= read -r response < /dev/tty
+  printf "%s" "$response"
+}
+
 while getopts ":n:h" OPTION; do
   case "$OPTION" in
     n)
@@ -116,7 +134,8 @@ printf "Node user:        %s\n" "$username"
 printf "Database:         %s\n" "$databasename"
 printf "PM2 process:      %s\n\n" "$processname"
 
-read -r -p "WARNING: This will permanently replace database '$databasename'. Type \"yes\" to continue: " agreement
+require_interactive_tty
+agreement="$(read_from_tty "WARNING: This will permanently replace database '$databasename'. Type \"yes\" to continue: ")"
 if [[ "$agreement" != "yes" ]]; then
   printf "\nRepair cancelled.\n\n"
   exit 1
--- a/tools/fix_node.sh
+++ b/tools/fix_node.sh
@@ -1,18 +1,28 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
-trap 'echo -e "\n[ERROR] Line $LINENO failed. Aborting.\n\n" >&2' ERR
 
-# Defaults for mainnet
+on_error() {
+  local exit_status=$?
+  printf "\n[ERROR] Line %s failed. Repair stopped.\n\n" "${BASH_LINENO[0]}" >&2
+  exit "$exit_status"
+}
+trap on_error ERR
+
+readonly TOOL_VERSION="1.4.0"
+
 network="mainnet"
 username="adamant"
 databasename="adamant_main"
 configfile="config.json"
 processname="adamant"
-port="36666" # Default mainnet REST port; later re-read from config
+port="36666"
 image_url="https://explorer.adamant.im/db_backup.sql.gz"
 
-# Options
-while getopts ":n:" OPTION; do
+usage() {
+  printf "Usage: %s [-n mainnet|testnet]\n" "${0##*/}"
+}
+
+while getopts ":n:h" OPTION; do
   case "$OPTION" in
     n)
       case "$OPTARG" in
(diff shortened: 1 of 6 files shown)


metalisk:
@claude /review

tools/fix_node.sh
```diff
@@ -22,200 +32,231 @@ while getopts ":n:" OPTION; do
...
 
-# Heavy lifting under node user
-NODE_HOME="$(eval echo "~$username")"
-REPO_DIR="${NODE_HOME}/adamant"
+# shellcheck disable=SC2016
+runuser -u "$username" -- env HOME="$NODE_HOME" PROCESS_NAME="$processname" bash -c '
+  set -Eeuo pipefail
+  source "$HOME/.nvm/nvm.sh"
+  command -v pm2 >/dev/null
+  pm2 describe "$PROCESS_NAME" >/dev/null
```
metalisk: 🟠 **Correctness — preflight aborts the exact disaster-recovery case this script exists for**

```bash
pm2 describe "$PROCESS_NAME" >/dev/null
```

This runs as a bare command under `set -Eeuo pipefail`. `pm2 describe` exits non-zero when the process isn't registered with pm2 — which is precisely the state you'd expect after a crash, a lost pm2 dump file, or an incomplete prior install. The `on_error` trap fires and the repair aborts before any download or destructive action, in the scenario the tool is meant to handle.

`install_node.sh` already gets this right at the equivalent spot:
```bash
if pm2 describe "$PROCESS_NAME" >/dev/null 2>&1; then
  pm2 restart "$PROCESS_NAME" --update-env
else
  ...
fi
```

Suggest: make the process-presence check a warning, not a hard failure, and guard the later `pm2 stop`/`pm2 restart` calls the same way `install_node.sh` does.

tools/fix_node.sh
```diff
@@ -22,200 +32,231 @@ while getopts ":n:" OPTION; do
...
-  echo
-  if [[ "\$network" == "mainnet" ]]; then
-    pm2 start --name "\$processname" app.js
-  else
-    pm2 start --name "\$processname" app.js -- --config "test/config.json" --genesis "test/genesisBlock.json"
-  fi
-fi
-pm2 save || true
+cd "$REPO_DIR"
+gunzip "$IMAGE_FILENAME"
```
metalisk: 🟡 **Low — `gunzip` here has no `-f` and no preceding `rm -f`, unlike the first EOSU block**

The download block (line 190) does `rm -f "${IMAGE_FILENAME}.part" "$IMAGE_FILENAME" "$IMAGE_UNZIPPED_FILENAME"` before fetching. This second block extracts without that same cleanup:
```bash
gunzip "$IMAGE_FILENAME"
```
If a stale `$IMAGE_UNZIPPED_FILENAME` exists from an interrupted prior run, `gunzip` exits non-zero ("already exists -- skipping") *after* the database has already been dropped/recreated a few lines up — leaving an empty DB and no clean way to recover except re-running. Mirror the cleanup here: `rm -f "$IMAGE_UNZIPPED_FILENAME"` immediately before the `gunzip` call (or add `-f`).

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
-  wget --progress=bar:force:noscroll "$image_url" -O "$image_filename" 2>/dev/tty
-  printf "\nUnzipping the blockchain image (may take a few minutes)…\n\n"
-  gunzip -f "$image_filename"
-  printf "\nLoading the image into '$databasename' database…\n\n"
-  psql "$databasename" < "$image_unzipped_filename"
-  printf "\nCleaning up temp image…\n"
-  rm -f "$image_unzipped_filename"
+config_tmp="$(mktemp "${CONFIG_FILE}.tmp.XXXXXX")"
+jq '.db.password = env.DB_PASSWORD_DECODED' "$CONFIG_FILE" > "$config_tmp"
+chmod --reference="$CONFIG_FILE" "$config_tmp" 2>/dev/null || true
```
metalisk: 🟡 **Low — TOCTOU: `chmod --reference` re-widens the temp file that holds the plaintext DB password**

```bash
config_tmp="$(mktemp "${CONFIG_FILE}.tmp.XXXXXX")"   # 0600, correct
jq '.db.password = env.DB_PASSWORD_DECODED' "$CONFIG_FILE" > "$config_tmp"  # still 0600
chmod --reference="$CONFIG_FILE" "$config_tmp" 2>/dev/null || true          # config.default.json is 0644 -> now 0644
mv "$config_tmp" "$CONFIG_FILE"
```
`config.default.json` ships at `0644`, and `cp` preserves that mode, so on a fresh install `$CONFIG_FILE` is `0644`. The `chmod --reference` call copies that mode onto the temp file *while it still contains the plaintext password*, briefly making it world-readable before the `mv`. Exploiting it requires a second local user racing a random `mktemp` suffix within a single `jq`+`mv` window, so this is low severity on a dedicated VPS — but it's a one-line fix: drop `chmod --reference` and just `chmod 0600 "$config_tmp"` (or rely on mktemp's default 0600 and never widen it).

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
 else
-  pm2 start --name adamanttest app.js -- --config test/config.json --genesis test/genesisBlock.json
+  printf "Configuration file '%s' already exists; preserving its settings.\n" "$CONFIG_FILE"
+fi
+
+DB_PASSWORD_DECODED="$(printf '%s' "$DB_PASSWORD_BASE64" | base64 --decode)"
+export DB_PASSWORD_DECODED
+config_tmp="$(mktemp "${CONFIG_FILE}.tmp.XXXXXX")"
+jq '.db.password = env.DB_PASSWORD_DECODED' "$CONFIG_FILE" > "$config_tmp"
+chmod --reference="$CONFIG_FILE" "$config_tmp" 2>/dev/null || true
```
metalisk: 🟡 **Low — same TOCTOU pattern as `install_node.sh`**

Identical `mktemp` → `jq` → `chmod --reference` → `mv` sequence widens the password-containing temp file from `0600` to `0644` before the `mv`. Same fix applies here: drop `chmod --reference` and keep the temp file at `0600`.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
+  exit 1
+fi
+REPO_DIR="${NODE_HOME}/adamant"
+
+# Create or update the PostgreSQL role and database without masking SQL errors.
+role_exists="$(runuser -u postgres -- psql -XAtqc \
+  "SELECT 1 FROM pg_roles WHERE rolname = '${username}'" postgres)"
+if [[ "$role_exists" == "1" ]]; then
+  printf "\nUpdating the password for existing PostgreSQL role '%s'.\n" "$username"
+  runuser -u postgres -- psql -Xv ON_ERROR_STOP=1 -c \
```
metalisk: 🟠 **Info/low but concretely exploitable — `-v` (verbose) on a password-setting `psql` call writes the plaintext password to the persistent install log**

```bash
runuser -u postgres -- psql -Xv ON_ERROR_STOP=1 -c \
    "ALTER ROLE ${username} WITH LOGIN PASSWORD '${DB_PASSWORD_SQL}';" postgres
```
All stdout/stderr is captured via `exec > >(tee -a "$LOGFILE")` near the top of the script, and `psql -v` (here used as a flag, distinct from `-v NAME=VALUE`) echoes back the statement it's about to run — including the literal password — to stdout. That gets written permanently to `/var/log/adamant_<network>_install.log`. This is a more durable exposure than the in-memory `DB_PASSWORD_SQL` variable lifetime: the log persists after install and may end up in backups/support bundles.

Same pattern at line 309 (`CREATE ROLE`) and the equivalent in `install_node_centos.sh` (lines 297, 301).

Suggest dropping `-v` (keep `-X` to skip `.psqlrc`) for the two password-setting calls specifically: `psql -X ON_ERROR_STOP=1 -c ...`. The other `psql -Xv` calls that don't touch secrets are fine as-is.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
+  exit 1
+fi
+PG_BIN_DIR="$(dirname "$PSQL_BIN")"
+
+role_exists="$(runuser -u postgres -- "$PSQL_BIN" -XAtqc \
+  "SELECT 1 FROM pg_roles WHERE rolname = '${username}'" postgres)"
+if [[ "$role_exists" == "1" ]]; then
+  printf "\nUpdating the password for existing PostgreSQL role '%s'.\n" "$username"
+  runuser -u postgres -- "$PSQL_BIN" -Xv ON_ERROR_STOP=1 -c \
+    "ALTER ROLE ${username} WITH LOGIN PASSWORD '${DB_PASSWORD_SQL}';" postgres
```
metalisk: 🟠 Same issue as `install_node.sh:305` — `-v` here echoes the `ALTER ROLE ... PASSWORD '...'` statement (plaintext password included) into the install log via the `tee` redirection set up earlier in the script. Recommend dropping `-v` for this call and the `CREATE ROLE` call at line 302.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
+architecture="$(dpkg --print-architecture)"
+pgdg_release_url="${pgdg_url}/dists/${VERSION_CODENAME}-pgdg/Release"
+pgdg_enabled=false
+postgresql_package="postgresql"
+
+case "$architecture" in
+  amd64|arm64|ppc64el)
+    if curl -fsSL -o /dev/null "$pgdg_release_url"; then
+      install -d -m 0755 /usr/share/postgresql-common/pgdg
+      curl -fsSL https://www.postgresql.org/media/keys/ACCC4CF8.asc \
```
metalisk, line 273: 🟡 **Medium — PostgreSQL APT signing key fetched at runtime with no fingerprint check**

```bash
curl -fsSL https://www.postgresql.org/media/keys/ACCC4CF8.asc \
  -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc
```
This key becomes the APT `Signed-By` trust anchor for everything PostgreSQL-related installed afterward, but its content is never checked against the known PGDG key fingerprint (`B97B 0AFC AA1A 47F0 44F2  44A0 7FCC 7D46 ACCC 4CF8`). HTTPS covers passive MITM, but not a compromise of postgresql.org's static file hosting — at which point this becomes the trust root for malicious PostgreSQL packages. This mirrors the documented upstream PGDG bootstrap, so it's not a regression, but since the script already goes out of its way to validate other downloads (`gzip --test` on the snapshot, `git check-ref-format` on the branch), it'd be consistent to also verify the key's fingerprint, or simply embed the (rarely-changing) key as a heredoc in the script to remove the runtime fetch entirely.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
+config_tmp="$(mktemp "${CONFIG_FILE}.tmp.XXXXXX")"
+jq '.db.password = env.DB_PASSWORD_DECODED' "$CONFIG_FILE" > "$config_tmp"
+chmod --reference="$CONFIG_FILE" "$config_tmp" 2>/dev/null || true
+mv "$config_tmp" "$CONFIG_FILE"
+unset DB_PASSWORD_DECODED DB_PASSWORD_BASE64
+
+if [[ "$USE_IMAGE" == "true" ]]; then
+  printf "\nDownloading the %s blockchain image.\n" "$NETWORK"
+  rm -f "${IMAGE_FILENAME}.part" "$IMAGE_FILENAME" "$IMAGE_UNZIPPED_FILENAME"
+  wget --progress=dot:giga "$IMAGE_URL" -O "${IMAGE_FILENAME}.part"
```
metalisk, line 575: 🟡 **Medium — blockchain snapshot integrity is checked structurally (`gzip --test`) but not by content hash**

Same pattern exists in `fix_node.sh` and `install_node_centos.sh`. A structurally valid gzip from a compromised `explorer.adamant.im` would pass the only check performed and then get loaded into a freshly-created (or, in `fix_node.sh`, freshly-*dropped*) database via `psql ... < file`. Since `psql` honors backslash meta-commands (e.g. `\!`) when reading a script file, a malicious dump could also achieve command execution as the node user during import — not just data corruption.

Worth publishing a `.sha256` sidecar (ideally from an origin independent of the snapshot host, e.g. GitHub Releases) and verifying it before the `gzip --test`/extract/load sequence. Not a blocker — the attack requires origin compromise — but cheap to add given the snapshot URL is already a fixed constant.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
-cd /tmp || exit 1
-sudo -u postgres psql -c "CREATE ROLE ${username} LOGIN PASSWORD '${DB_PASSWORD_SQL}';" || true
-sudo -u postgres psql -c "CREATE DATABASE ${databasename} OWNER ${username};" || true
-sudo -u postgres psql -c "ALTER DATABASE ${databasename} OWNER TO ${username};" || true
-
-# ------- Commands below run as the ADM node user; Variables expanded by parent -------
-su - "$username" <<EOSU
+printf "\nUpdating installed system packages. Existing package configuration files will be preserved.\n"
+apt-get update
+apt-get "${APT_OPTIONS[@]}" upgrade
```
metalisk, line 306: 🟡 **Medium — unconditional `apt-get upgrade` on a script that explicitly supports re-runs against existing installs**

The old confirmation prompt read "Intended for NEW droplets. Existing data MAY BE DAMAGED"; the new one (`The script will update packages and configure ADAMANT services`) is accurate but softer, while the script now deliberately supports running against hosts that already have ADAMANT, PostgreSQL, and pm2 set up. Combined with `NEEDRESTART_MODE=a` (auto-restart affected services with no prompt), running this against a host that also runs other services could silently bounce unrelated daemons. Consider a stronger explicit warning at the confirmation prompt (or detecting an existing install and skipping the full-system upgrade in favor of just the ADAMANT-specific packages) since the idempotent re-run path is now a first-class, encouraged use case.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
+printf "\nInstalling package-management and build prerequisites.\n"
+dnf -y install \
+  ca-certificates curl git gzip jq tar wget \
+  gcc gcc-c++ make automake autoconf libtool pkgconf-pkg-config python3
+
+# Use the official PGDG repository on fresh installations. Installing the
+# repository is harmless on reruns and keeps installed PostgreSQL packages
+# eligible for security updates.
+pgdg_rpm="https://download.postgresql.org/pub/repos/yum/reporpms/EL-${el_major}-${pgdg_arch}/pgdg-redhat-repo-latest.noarch.rpm"
+dnf -y install "$pgdg_rpm"
```
metalisk: 🟡 **Low/medium — no probe or fallback if the PGDG repo RPM is unreachable**

```bash
dnf -y install "$pgdg_rpm"
```
Unlike `install_node.sh`, which probes `pgdg_release_url` first and falls back to distro packages on failure, this installs directly from a `-latest` URL with no fallback. Under `set -Eeuo pipefail` any network blip here aborts the whole installer — *after* the operator has already entered and confirmed their DB password — forcing a full re-run including re-entering the password. Consider mirroring the Ubuntu script's probe-then-fallback pattern.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
-printf "Check your node status with the command: 'pm2 show %s'\n" "$processname"
-printf "To check the current node height, run: 'curl http://localhost:%s/api/blocks/getHeight'\n" "$port"
-printf "Thank you for supporting the truly decentralized ADAMANT Messenger.\n\n"
-su - "$username"
+unset DB_PASSWORD_BASE64 DB_PASSWORD_SQL
+
+# shellcheck disable=SC2016
+NODE_BIN_DIR="$(runuser -u "$username" -- env HOME="$NODE_HOME" bash -c \
+  'source "$HOME/.nvm/nvm.sh" && dirname "$(command -v node)"')"
+printf "\nEnabling pm2 startup for user '%s'.\n" "$username"
```
metalisk: 🟢 **Low — `pm2 startup systemd` here has no systemd-presence guard, unlike the Ubuntu script**

`install_node.sh` guards the equivalent block with `if command -v systemctl >/dev/null 2>&1 && [[ "$(ps -p 1 -o comm=)" == "systemd" ]]`, with a fallback warning otherwise. This block runs unconditionally. In practice the earlier unconditional `systemctl enable --now "$postgresql_service"` (line 263) would already abort the script on a non-systemd host before reaching here, so this is currently harmless — but it's a maintenance trap if that earlier call is ever made conditional. Worth copying the same guard for consistency.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
+if dpkg-query -W -f='${Package}\n' 'postgresql-[0-9]*' 2>/dev/null \
+  | grep -Eq '^postgresql-[0-9]+$'; then
+  printf "Existing PostgreSQL server installation detected; preserving its major version.\n"
+else
+  printf "Installing PostgreSQL server package '%s'.\n" "$postgresql_package"
+  apt-get "${APT_OPTIONS[@]}" install "$postgresql_package"
+fi
+
+if command -v systemctl >/dev/null 2>&1 && systemctl list-unit-files postgresql.service >/dev/null 2>&1; then
+  systemctl enable --now postgresql
```
metalisk: 🟢 **Low — no readiness poll between starting PostgreSQL and the first `psql` call**

```bash
systemctl enable --now postgresql
...
runuser -u postgres -- psql ...   # could race startup
```
On a freshly-initialized cluster this window is usually negligible, and a race here fails loudly via `set -Eeuo pipefail` rather than silently — so this is a minor robustness nit, not a real risk. A short `pg_isready` poll loop before the first `psql -c` would remove the (rare) need to re-run the installer after a race-induced failure. Same applies in `install_node_centos.sh` around line 263.

metalisk (commented):
## Ultra Code Review: PR #241

Multi-agent review (4 independent dimension passes + adversarial verification on every finding) of the installer modernization. Overall this is solid work: passwords cross privilege boundaries via base64 rather than raw interpolation, destructive DB operations are sequenced strictly after download/integrity checks, `set -Eeuo pipefail` + an ERR trap make failures loud instead of silent, and the CentOS script gains testnet image parity with Ubuntu.

No finding survived adversarial review at "critical" or even confidently at "high" — most security-flavored findings (PGDG key pinning, nvm curl-pipe-bash, snapshot checksums) are real supply-chain hardening gaps but require either origin-server compromise or are already-accepted patterns from upstream docs, so they land at medium/low. The findings worth fixing before merge are the two real correctness bugs (`fix_node.sh` pm2 preflight, `chmod --reference` TOCTOU) and one concrete information-disclosure regression (`psql -v` echoing the plaintext password into the persistent install log). Inline comments below.

**Verdict: Approve with comments** — nothing here blocks merge, but the three items flagged 🟠 are worth a follow-up commit since they're cheap fixes with concrete impact.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
 port="36666"
-nodejs="jod" # LTS=22 by default
+nodejs="24"
 image_url="https://explorer.adamant.im/db_backup.sql.gz"
 
-# Parse options
-while getopts ":b:n:j:" OPTION; do
+usage() {
+  printf "Usage: %s [-b branch] [-n mainnet|testnet] [-j 22|24|26]\n" "${0##*/}"
+}
```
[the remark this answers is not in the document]
metalisk: Fixed locally: `usage()` now advertises `-h` and documents the accepted Node.js aliases (`jod=22`, `krypton=24`) plus the default.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
 port="36666"
-nodejs="jod"
+nodejs="24"
+image_url="https://explorer.adamant.im/db_backup.sql.gz"
 
-while getopts 'b:n:j:' OPTION; do
-  OPTARG=$(echo "$OPTARG" | xargs)
+usage() {
+  printf "Usage: %s [-b branch] [-n mainnet|testnet] [-j 22|24|26]\n" "${0##*/}"
+}
```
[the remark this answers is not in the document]
metalisk: Fixed locally: `usage()` now includes `-h` and the Node.js alias/default note, matching the implemented option parser.

tools/fix_node.sh
```diff
@@ -1,18 +1,28 @@
...
 processname="adamant"
-port="36666" # Default mainnet REST port; later re-read from config
+port="36666"
 image_url="https://explorer.adamant.im/db_backup.sql.gz"
 
-# Options
-while getopts ":n:" OPTION; do
+usage() {
+  printf "Usage: %s [-n mainnet|testnet]\n" "${0##*/}"
+}
```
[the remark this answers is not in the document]
metalisk: Fixed locally: `usage()` now includes the supported `-h` flag.

tools/install_node.sh
```diff
@@ -9,265 +18,464 @@ databasename="adamant_main"
...
-npm i -g pm2
-
-# pm2-logrotate
-printf "\n\n"
-pm2 install pm2-logrotate
+printf "\nInstalling or updating nvm v%s and Node.js %s.\n" "$NVM_INSTALL_VERSION" "$NODEJS_VERSION"
+export NVM_DIR="$HOME/.nvm"
+curl -fsSL "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_INSTALL_VERSION}/install.sh" | bash
+# shellcheck disable=SC1091
+source "$NVM_DIR/nvm.sh"
```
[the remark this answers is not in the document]
metalisk: Fixed locally: nvm is now installed/updated from the pinned Git tag via `git clone`/`git fetch` + checkout instead of `curl | bash`; the shell profile loader is appended explicitly when missing.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,440 @@ databasename="adamant_main"
...
+  IMAGE_UNZIPPED_FILENAME="$image_unzipped_filename" \
+  bash <<'EOSU'
+set -Eeuo pipefail
+trap 'status=$?; printf "\n[ERROR] User setup failed at line %s.\n\n" "$LINENO" >&2; exit "$status"' ERR
+
+printf "\nInstalling or updating nvm v%s and Node.js %s.\n" "$NVM_INSTALL_VERSION" "$NODEJS_VERSION"
+export NVM_DIR="$HOME/.nvm"
+curl -fsSL "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_INSTALL_VERSION}/install.sh" | bash
+# shellcheck disable=SC1091
+source "$NVM_DIR/nvm.sh"
```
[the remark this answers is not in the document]
metalisk: Fixed locally: same nvm change as the Ubuntu installer, using the pinned Git tag instead of executing the remote installer through `curl | bash`.

tools/fix_node.sh
metalisk: Fixed locally: the repair preflight now only requires `pm2` to exist; a missing process is a warning, stop is conditional, and restart falls back to `pm2 start` with the correct mainnet/testnet arguments.

tools/fix_node.sh
metalisk: Fixed locally: the repair load step removes any stale unzipped dump before extraction and uses `gunzip -f`, so a prior interrupted run cannot leave an empty recreated DB because of an existing output file.

tools/install_node.sh
metalisk: Fixed locally: the config temp file now stays `0600` (`chmod 0600`) instead of copying a potentially world-readable mode from the existing config.

tools/install_node_centos.sh
metalisk: Fixed locally: same config temp-file mode fix in the EL installer; the password-containing temp file is kept at `0600` until `mv`.

tools/install_node.sh
metalisk: Fixed locally: password-setting SQL is now sent to `psql` via stdin with `--set=ON_ERROR_STOP=1`, avoiding the password in the process argv and avoiding the ambiguous `-Xv` form. Other `psql` calls were also normalized to `--set=ON_ERROR_STOP=1`.

tools/install_node_centos.sh
metalisk: Fixed locally: the EL installer now uses stdin for the password-setting SQL and `--set=ON_ERROR_STOP=1`, matching the Ubuntu fix.

tools/install_node.sh
metalisk: Fixed locally: after downloading the PGDG APT key, the script verifies the expected fingerprint `B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8` before writing the repo source.

tools/install_node.sh
metalisk: Not fixed in this PR: I agree content-hash validation would be stronger, but it needs a published trusted checksum/signature sidecar for each moving snapshot. Without that external artifact, the script cannot verify content authenticity beyond transport + gzip structure. I would track this as a follow-up infrastructure/docs task.

tools/install_node.sh
metalisk: Fixed locally: the installer now explicitly warns before confirmation that package upgrades may restart PostgreSQL, Redis, and other affected system services.

tools/install_node_centos.sh
metalisk: Fixed locally: the EL installer now probes the PGDG RPM first; if unavailable, it warns and falls back to distribution PostgreSQL packages instead of aborting immediately.

tools/install_node_centos.sh
metalisk: Fixed locally: copied the Ubuntu-style systemd guard around `pm2 startup`; non-systemd hosts now get a warning instead of an unconditional systemd startup call.

tools/install_node.sh
metalisk: Fixed locally: added a short `pg_isready` poll after PostgreSQL start in both installers before the first `psql` operation.

metalisk:
Follow-up from the current active `gh` account (`metalisk`):

Addressed the new unresolved review threads locally, without committing:

- added `-h` to all usage strings and documented Node.js aliases in both installers;
- replaced `curl | bash` nvm setup with pinned Git tag checkout/update;
- made `fix_node.sh` tolerate a missing PM2 process and recreate it after repair;
- hardened repair extraction against stale unzipped dump files;
- kept DB-password config temp files at `0600`;
- moved password-setting SQL to `psql` stdin and normalized `ON_ERROR_STOP` usage;
- added PGDG APT key fingerprint verification;
- strengthened warnings around package upgrades and service restarts;
- added EL PGDG probe/fallback, systemd guard for PM2 startup, and PostgreSQL readiness polling.

One thread was intentionally not changed in this PR: blockchain snapshot content-hash verification needs a trusted published checksum/signature sidecar for each moving snapshot. The scripts cannot verify that authenticity source locally until the release/snapshot infrastructure provides it, so I replied there as a follow-up infrastructure item.

Validation run after the local fixes:

```sh
bash -n tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh
npx --yes shellcheck@4.1.0 tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh
git diff --check -- tools/install_node.sh tools/install_node_centos.sh tools/fix_node.sh README.md
bash tools/install_node.sh -h
bash tools/install_node_centos.sh -h
bash tools/fix_node.sh -h
```

tools/fix_node.sh
```diff
@@ -22,200 +59,292 @@ while getopts ":n:" OPTION; do
...
-pm2 save || true
+  pm2 save
+'
+
+printf "\nTerminating active connections to database '%s'.\n" "$databasename"
+runuser -u postgres -- psql -X --set=ON_ERROR_STOP=1 -c \
+  "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='${databasename}' AND pid <> pg_backend_pid();" \
+  postgres
+
+printf "Dropping database '%s' to free disk space before downloading the image.\n" "$databasename"
```
metalisk, line 277: 🟡 **Behavioral regression (now documented, but worth confirming intentional): database is dropped *before* the snapshot is downloaded and validated**

The previous version of this script deliberately downloaded and `gzip --test`-validated the image *before* touching the database, specifically so a failed download or corrupt snapshot would leave the existing (stale but intact) database alone. This round reorders it — drop happens first, "to free disk space before downloading the image" (per the comment and the updated confirmation prompt at the `agreement` read).

This is a real tradeoff: it solves a legitimate problem (insufficient disk for both old DB + new compressed/decompressed snapshot side-by-side) but reintroduces the failure mode the original ordering was designed to avoid — if `wget` or `gzip --test` fails after this point, the node is left with **no database at all** until a successful re-run, rather than the previous stale-but-running state.

Given the prompt text was already updated to say "This will stop the node and drop database... before downloading the replacement image," this looks like a conscious decision rather than an oversight — flagging mainly to confirm that's the case, and to suggest a disk-space preflight check (`df`) before the drop so the operator isn't surprised by a download failure on a half-deleted node when the actual cause is insufficient space for the new image.

metalisk (commented):
## Follow-up review: PR #241 (re-reviewed after fixes)

Great turnaround — every item from the previous review round was addressed correctly:

- ✅ `fix_node.sh` pm2 preflight now warns instead of aborting when the process isn't registered
- ✅ `chmod --reference` TOCTOU replaced with explicit `chmod 0600` on the config temp file (both installers)
- ✅ `psql -v`/`-Xv` removed from password-setting calls — now `psql -X --set=ON_ERROR_STOP=1 ... <<EOSQL` heredocs, so the plaintext password is no longer echoed into the persistent install log
- ✅ PGDG signing key fingerprint is now verified against the known PGDG fingerprint before being trusted
- ✅ CentOS PGDG RPM install now probes the URL first and falls back to distro packages, matching the Ubuntu script
- ✅ `pg_isready` polling added after starting PostgreSQL in both installers
- ✅ `apt-get upgrade` confirmation prompt now explicitly warns that package upgrades may restart PostgreSQL/Redis/other services
- ✅ nvm is now installed via a pinned `git clone --branch vX.Y.Z` instead of `curl | bash`, which is a meaningfully better posture than executing an arbitrary downloaded script (no signature, but at least an addressable, immutable-by-convention tag rather than blind script execution)
- ✅ `gunzip`'s stale-file failure mode in `fix_node.sh` is gone entirely — extraction is now a streaming `gzip -dc | psql` pipeline with no intermediate `.sql` file to collide with

Two new things worth a look from the changes added in this round (mostly `fix: add user permissions repair`, `enhance PostgreSQL service detection`, and `interactive terminal requirements` commits):

tools/install_node.sh
```diff
@@ -9,265 +18,604 @@ databasename="adamant_main"
...
-  printf "User '%s' has been created.\n\n" "$username"
+DB_PASSWORD_BASE64="$(printf '%s' "$DB_PASSWORD" | base64 | tr -d '\n')"
+unset DB_PASSWORD
+
+printf "\nUpdating package indexes and installing package-management prerequisites.\n"
+export DEBIAN_FRONTEND=noninteractive
+export NEEDRESTART_MODE=a
+# Skip needrestart hooks during installer-managed apt runs. Some VPS images ship
+# a broken needrestart.conf, which can print scary Perl parse errors after dpkg.
+export NEEDRESTART_SUSPEND=1
```
metalisk, line 252: 🟡 **`NEEDRESTART_SUSPEND=1` makes the adjacent `NEEDRESTART_MODE=a` a no-op, and silently disables automatic restart of services after library upgrades**

```bash
export NEEDRESTART_MODE=a
export NEEDRESTART_SUSPEND=1
```
`NEEDRESTART_SUSPEND` (when set to any value) causes needrestart's apt/dpkg hook to exit immediately without running at all — it doesn't just suppress the interactive prompt, it disables needrestart's restart logic entirely. So `NEEDRESTART_MODE=a` ("auto-restart affected services") is now dead configuration; it never gets a chance to act.

The comment explains the motivation (some VPS images ship a broken `needrestart.conf` that prints Perl parse errors), which is a legitimate problem to dodge. But the side effect is that services running on libraries upgraded by `apt-get upgrade` — including PostgreSQL itself, which the script restarts only if it isn't already running — won't get bounced automatically, and there's no warning to the operator that a manual restart/reboot may be needed for upgraded libraries to take effect. This is the same pattern in `fix_node.sh` at the equivalent spot.

Worth either: dropping `NEEDRESTART_MODE=a` since it no longer does anything, or replacing the pair with something that avoids the broken-conf crash without fully disabling restart awareness, e.g. wrapping the `needrestart` invocation/config check instead of suspending it globally — or at minimum, printing a note that a reboot may be required after the upgrade since automatic service restarts are disabled.

metalisk: Thanks, it's intentional.

tools/fix_node.sh
metalisk: Thanks, it's intentional.

tools/install_node.sh
```diff
@@ -9,265 +18,604 @@ databasename="adamant_main"
...
+  NETWORK="$network" \
+  BRANCH="$branch" \
+  NODEJS_VERSION="$nodejs" \
+  NVM_INSTALL_VERSION="$NVM_VERSION" \
+  REPO_DIR="$REPO_DIR" \
+  CONFIG_FILE="$configfile" \
+  PROCESS_NAME="$processname" \
+  DATABASE_NAME="$databasename" \
+  DB_PASSWORD_BASE64="$DB_PASSWORD_BASE64" \
+  USE_IMAGE="$IMAGE" \
```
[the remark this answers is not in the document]
metalisk: Fixed locally, not committed yet.

The installer no longer passes the database password through `env DB_PASSWORD_BASE64=...` in the `runuser` argv. It now writes the password to a random temp file in the node user's home directory with mode `0600`, changes ownership to the node user, passes only the temp file path as `DB_PASSWORD_FILE`, reads the password from that protected file inside the user setup block, and removes the file on both normal completion and installer errors.

Version bumped to `2.4.8` for this fix.

tools/install_node_centos.sh
```diff
@@ -7,216 +19,566 @@ databasename="adamant_main"
...
+  PATH="${PG_BIN_DIR}:/usr/local/bin:/usr/bin:/bin" \
+  NETWORK="$network" \
+  BRANCH="$branch" \
+  NODEJS_VERSION="$nodejs" \
+  NVM_INSTALL_VERSION="$NVM_VERSION" \
+  REPO_DIR="$REPO_DIR" \
+  CONFIG_FILE="$configfile" \
+  PROCESS_NAME="$processname" \
+  DATABASE_NAME="$databasename" \
+  DB_PASSWORD_BASE64="$DB_PASSWORD_BASE64" \
```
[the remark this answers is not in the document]
metalisk: Fixed locally, not committed yet.

The CentOS/RHEL installer now uses the same safer handoff as the Ubuntu installer: the database password is written to a random temp file in the node user's home directory with mode `0600`, owned by the node user, and only the temp file path is passed through `DB_PASSWORD_FILE`. The file is cleaned up on normal completion and from the root-level error handler, so the secret is no longer exposed in `ps` argv.

Version bumped to `2.4.6` for this fix.

adamantmm (approved):

PR #249, opened 2026-06-29
Fix: Sync watchdog so a stalled sync cannot wedge the node

### LINKED ISSUE #248: [Bug] Stalled sync wedges the node: syncing() stays true, all live blocks rejected, no recovery
## Summary

A syncing node can wedge permanently: it stops at a height, rejects every live block with `Client not yet ready to receive block`, never starts another sync, and only a manual restart recovers it. The chain it built is valid — this is a **liveness/recovery defect in the sync state machine, not a consensus break**.

Observed on a fresh `dev` build syncing from height 0 (node frozen at height `26,420,804` for 3+ days, `pm2` `online`, `0` restarts, CPU ~0%).

## Root cause

A single `__private.sync` run (`modules/loader.js`) is driven by callback-based `async` flows. If any inner step never invokes its callback (e.g. a swallowed callback in the block-processing path — cf. prior bug `fix: onReceiveBlock doesn't call callback`), the `async.series` completion handler never runs. That handler is the only place that clears the sync state, so:

1. `__private.syncIntervalId` stays set ⇒ `loader.syncing()` stays `true` forever.
2. `isReadyToReceiveBlock()` in `modules/blocks/process.js` is `false` while `syncing()` ⇒ **every** incoming live block is discarded (`Client not yet ready to receive block`).
3. `__private.syncTimer` only starts a new sync when `!syncing()` ⇒ **no recovery sync is ever started**.
4. The `sequence.add` task that wraps the sync never completes, blocking the main `library.sequence` queue.

Net effect: a node that is connected to peers but frozen indefinitely, even though its chain is correct.

### Evidence

- `loader/sync` trace `Sync trigger` fires every second for days, with **zero** `Starting sync` / `Finished sync` in the same window → a single sync started and never finished.
- `Client not yet ready to receive block` logged continuously (tens of thousands of times per log file).
- Local tip block matches the canonical chain on public nodes byte-for-byte (id, previousBlock, payloadHash, generator, signature) → no fork, no validation failure.
- No stuck DB query; event loop idle → blocked on an in-process callback that never fires.

## Expected behavior

A sync run that makes no progress for a bounded period must be aborted and the loader must recover automatically (release `syncing()`, accept live blocks, and start a fresh sync) without operator intervention. Block validation and chain state must be untouched.

## Proposed fix

Add a **progress-based sync watchdog**: while a sync is active, sample blockchain height; if no new block is applied within a bounded window, abort the stalled run exactly once (release sync state, signal the in-flight series to bail at its next safe checkpoint, refresh the peer cache) so the sync timer starts a fresh attempt. A slow-but-progressing sync is never aborted because the check is "did height advance within the window". Liveness-only; no change to block acceptance, serialization, IDs, or signatures.

## Scope / risk

- Touches `modules/loader.js` (sync orchestration) + new `helpers/syncWatchdog.js` + one operational constant.
- No consensus, serialization, or activation-height changes.
- Tests: watchdog helper unit tests + `isReadyToReceiveBlock` gate coverage in `modules/blocks/process.js`.

Related PR: _to be linked_

massivedev0:
## Description

A syncing node can wedge permanently: it freezes at a height, rejects every incoming live block with `Client not yet ready to receive block`, never starts another sync, and only a manual restart recovers it. The chain it built is valid — this is a **liveness/recovery defect in the sync state machine, not a consensus break**.

### Root cause

A single `__private.sync` run (`modules/loader.js`) is driven by callback-based `async` flows. If any inner step never invokes its callback (e.g. a swallowed callback in the block-processing path), the `async.series` completion handler never runs. That handler is the only place that clears the sync state, so:

1. `__private.syncIntervalId` stays set ⇒ `loader.syncing()` stays `true` forever.
2. `isReadyToReceiveBlock()` (`modules/blocks/process.js`) is `false` while `syncing()` ⇒ **every** live block is discarded.
3. `__private.syncTimer` only starts a new sync when `!syncing()` ⇒ no recovery sync is ever started.
4. The `sequence.add` wrapper task never completes, blocking the main `library.sequence` queue.

Diagnosed from a `dev` build syncing from height 0 that was frozen at height `26,420,804` for 3+ days: `Sync trigger` trace firing every second with zero `Starting sync`/`Finished sync`, `Client not yet ready to receive block` logged continuously, local tip matching the canonical chain byte-for-byte on public nodes (no fork), no stuck DB query, idle event loop.

### Fix

Add a **progress-based sync watchdog** (`helpers/syncWatchdog.js`). While a sync is active it samples blockchain height; if no new block is applied within a bounded window (`constants.syncStallTimeout`, default 5 min), it aborts the stalled run. The check is *"did height advance within the window"*, so a slow-but-progressing sync (a node far behind that keeps applying blocks) is **never** aborted — only a genuinely stalled run is.

Recovery rests on a single invariant: **once a run is aborted it can never *begin* a new state mutation; teardown only waits for a mutation already in flight to drain.**

- **Run-scoped abort.** Each `__private.sync` run owns a local `aborted` flag and a `shouldStop()` predicate (distinct from the shutdown-only `stopRequested`), so a fresh run can never clear an older run's signal.
- **Block apply is gated at the boundary.** `shouldStop()` is threaded `loadBlocksFromNetwork → loadBlocksFromPeer → verify.processBlock`, and `processBlock()` checks it **immediately before `applyBlock()`** — the chain-state mutation. A block parked anywhere in verification (`checkExists` / slot validation / tx checks) bails before applying when it resumes. The fork-2 `undoUnconfirmed` write inside `checkTransaction()` is gated the same way.
- **Unconfirmed phases** (`undoUnconfirmedList` / `applyUnconfirmedList`) are not started after abort (`skipOnStop`).
- **Drain before teardown.** `finishStalled()` releases `loader.syncing()` only once `modules.blocks.isActive` (a block apply mid-write) and the run-local `mutatingState` (an unconfirmed phase mid-write) are both clear. Completion is funnelled through a single `settled`-guarded `finishSync()`, so the watchdog and the natural `async.series` completion are mutually exclusive and a late (abandoned) completion is a harmless no-op. The watchdog abort reports no error to `async.retry`, so recovery comes from the next sync-timer tick rather than an immediate in-sequence retry.

This is **liveness-only**: for non-sync callers (`onReceiveBlock`, `generateBlock`) `shouldStop` is absent, so there is no change to block acceptance, validation, serialization, IDs, signatures, slot timing, rewards, rounds, or activation gating.

## Related issue

Closes #248

## How to test

Targeted unit tests (no DB/Redis required for these suites):

```
npm run test:single -- \
  test/unit/helpers/syncWatchdog.js \
  test/unit/modules/blocks.process.js \
  test/unit/modules/blocks.verify.js
```

Coverage:
- `test/unit/helpers/syncWatchdog.js` — stall detection: fires on no progress, ignores a progressing sync, fires once, disabled at `timeoutMs <= 0`, `stop()` / idempotent `start()`.
- `test/unit/modules/blocks.verify.js` — the apply-boundary gate: `processBlock()` verifies but does not apply when `shouldStop()` is true (applies otherwise / when no predicate is given); the fork-2 `undoUnconfirmed` write is skipped when aborted, performed otherwise.
- `test/unit/modules/blocks.process.js` — `isReadyToReceiveBlock()` gate branches (syncing / not-ready-to-sync / round ticking) and that `loadBlocksFromPeer` forwards `shouldStop` to `verify.processBlock`.

Broader regression run (executed locally, PostgreSQL + Redis up):

```
npm run test:unit:fast    # 861 passing
```

Lint (repo uses flat `eslint.config.js`):

```
npx eslint helpers/syncWatchdog.js modules/loader.js helpers/constants.js \
  modules/blocks/process.js modules/blocks/verify.js \
  test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js \
  test/unit/modules/blocks.verify.js   # clean
```

What was **not** run: `npm run test:api` (requires a running testnet) and a full from-zero mainnet resync. The change is liveness-only and does not touch validation paths, but reviewers may want an API/peer pass before merge.

### Manual reproduction / verification

On a node that exhibits the freeze: `Sync trigger` repeats every second with no `Finished sync`, and `/api/loader/status/sync` reports `syncing: true` with a frozen `height`. With this change, a sync that makes no progress for `syncStallTimeout` logs `Sync watchdog: no block progress, aborting stalled sync to allow recovery`, `loader.syncing()` returns to `false`, live blocks are accepted again, and a fresh sync starts on the next timer tick.

## Risk areas

- **Consensus / protocol:** none — block validation, serialization, IDs, signatures, rounds, rewards, and activation heights are untouched.
- **Concurrency:** an aborted run cannot begin a new state mutation — block apply is gated immediately before `applyBlock()` (and the fork-2 `undoUnconfirmed`), and the unconfirmed phases are gated by `skipOnStop`. Teardown waits for any already-started mutation (`modules.blocks.isActive` / `mutatingState`) to drain. So releasing `loader.syncing()` cannot let an abandoned flow mutate state concurrently with a fresh run.
- **Tuning:** `constants.syncStallTimeout` (default 5 min) is deliberately generous; set to `0` to disable the watchdog.
- **Scope:** only `__private.sync` is guarded; the startup rebuild path (`loadBlockChain`) is out of scope. If a state-mutating phase is itself permanently hung mid-write, `finishStalled()` keeps waiting (it will not force teardown into a possible concurrent write) — that case needs operator intervention.
- **Follow-ups (not in this PR):** the probable root-cause hang is `checkExists` in `verify.processBlock` doing `db.query(...).then(...)` with no `.catch`; adding the rejection handler is a separate root-cause fix. This PR is the liveness safety net.

## Checklist

- [x] No security checks weakened (nethash/version/handshake untouched).
- [x] Schema/SQL unaffected (no data-shape changes).
- [x] Tests added (watchdog helper) and adjacent gate coverage extended; targeted + fast unit suites run.
- [x] Operator note: new `syncStallTimeout` constant (liveness only, not consensus); `0` disables.

### DIFF
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -743,22 +743,26 @@ __private.sync = function (cb) {
         blocksToSync: __private.blocksToSync,
         timeoutMs: __private.syncTimeout
       });
-      // Signal the in-flight series to bail at its next safe checkpoint, then
-      // release the sync state once block application is idle.
+      // Set the run's abort flag, then release the sync state once any mutation
+      // that was already in progress has drained.
       aborted = true;
       finishStalled();
     }
   });
 
-  // Releases the sync state after a watchdog abort, but only once no state
-  // mutation is in flight. Block application (`applyBlock()`) and the unconfirmed
-  // undo/apply phases write mem/account/round/pool state outside
-  // `library.sequence`, so advancing the outer sequence while any of them runs
-  // could let a new sync or live block mutate the same state concurrently. We
-  // therefore wait for both `modules.blocks.isActive` and the run-local
-  // `mutatingState` to clear before tearing down. If the in-flight series reaches
-  // a checkpoint first, it finishes the run itself and this becomes a no-op via
-  // the `settled` guard.
+  // Recovery rests on one invariant: once `aborted` is set, this run can never
+  // begin a new state mutation.
+  //   - New block apply is blocked because `shouldStop()` is threaded into
+  //     `processBlock()` and checked right before `applyBlock()`, so a block
+  //     parked anywhere in verification bails before mutating when it resumes.
+  //   - New unconfirmed undo/apply phases are blocked by `skipOnStop()`.
+  // That makes it safe to release `loader.syncing()` even while an abandoned
+  // flow is still parked. The only thing left to guard is a mutation that had
+  // *already started* before the abort: `applyBlock()` and the unconfirmed
+  // phases write mem/account/round/pool state outside `library.sequence`, so we
+  // wait for `modules.blocks.isActive` and the run-local `mutatingState` to
+  // clear before tearing down. If the in-flight series reaches a checkpoint
+  // first, it finishes the run itself and this becomes a no-op via `settled`.
   function finishStalled () {
     if (settled) {
       return;
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -26,10 +26,6 @@ __private.syncInterval = 10000;
 __private.retries = 5;
 __private.stopRequested = false;
 __private.shutdownRequested = false;
-// Set true when the watchdog aborts a stalled sync; signals the in-flight sync
-// to bail at its next safe checkpoint without disabling future syncs (unlike
-// `stopRequested`, which is reserved for shutdown).
-__private.abortSync = false;
 // Maximum time a single sync run may go without block progress before it is
 // considered stalled and aborted so the loader can recover.
 __private.syncTimeout = constants.syncStallTimeout;
@@ -127,14 +123,12 @@ __private.requestStop = function () {
 };
 
 /**
- * Checks whether in-flight loader work should stop at its next safe boundary.
- * True on shutdown (`stopRequested`) or when the watchdog aborted a stalled
- * sync (`abortSync`).
+ * Checks if shutdown was requested.
  * @private
  * @return {boolean}
  */
 __private.isStopRequested = function () {
-  return __private.stopRequested || __private.abortSync;
+  return __private.stopRequested;
 };
 
 /**
@@ -605,9 +599,14 @@ __private.loadBlockChain = function () {
  * @implements {modules.blocks.getCommonBlock}
  * @return {setImmediateCallback} cb, err
  */
-__private.loadBlocksFromNetwork = function (cb) {
+__private.loadBlocksFromNetwork = function (cb, shouldStop) {
   var errorCount = 0;
   var loaded = false;
+
+  // Per-run stop predicate. Defaults to the shutdown signal; the sync watchdog
+  // passes a run-scoped predicate so an aborted run keeps its own stop signal
+  // and a fresh run cannot accidentally clear it.
+  shouldStop = shouldStop || __private.isStopRequested;
 
   self.getNetwork(function (err, network) {
     if (err) {
@@ -615,7 +614,7 @@ __private.loadBlocksFromNetwork = function (cb) {
     } else {
       async.whilst(
           function (testCb) {
-            return testCb(null, !__private.isStopRequested() && !loaded && errorCount < 5);
+            return testCb(null, !shouldStop() && !loaded && errorCount < 5);
           },
           function (next) {
             var peer = network.peers[Math.floor(Math.random() * network.peers.length)];
@@ -632,7 +631,7 @@ __private.loadBlocksFromNetwork = function (cb) {
                 loaded = lastValidBlock.id === lastBlock.id;
                 lastValidBlock = lastBlock = null;
                 next();
-              }, __private.isStopRequested);
+              }, shouldStop);
             }
 
             function getCommonBlock (cb) {
@@ -697,15 +696,24 @@ __private.sync = function (cb) {
   });
   library.bus.message('syncStarted');
 
-  __private.abortSync = false;
   __private.isActive = true;
   __private.syncTrigger(true);
 
+  // Run-scoped abort flag. Kept local (not on `__private`) so a fresh sync run
+  // can never clear the stop signal of an earlier, still-alive run, and so it
+  // stays distinct from the shutdown-only `stopRequested`.
+  var aborted = false;
   // Completion is funnelled through finishSync() and guarded by `settled` so the
-  // watchdog and the natural async.series completion are mutually exclusive: a
-  // stalled sync that the watchdog aborts releases the sync state exactly once,
-  // and a later (abandoned) completion of the series becomes a harmless no-op.
+  // watchdog and the natural async.series completion are mutually exclusive: the
+  // sync state is released exactly once, and a later (abandoned) completion of
+  // the series becomes a harmless no-op.
   var settled = false;
+
+  // Shutdown or this run's watchdog abort both stop the in-flight series at its
+  // next safe checkpoint.
+  function shouldStop () {
+    return __private.stopRequested || aborted;
+  }
 
   var watchdog = createSyncWatchdog({
     timeoutMs: __private.syncTimeout,
@@ -718,12 +726,36 @@ __private.sync = function (cb) {
         blocksToSync: __private.blocksToSync,
         timeoutMs: __private.syncTimeout
       });
-      // Ask the in-flight series to bail at its next safe checkpoint, then
-      // release the sync state so the sync timer can start a fresh attempt.
-      __private.abortSync = true;
-      finishSync('Sync watchdog timeout: no block progress');
+      // Signal the in-flight series to bail at its next safe checkpoint, then
+      // release the sync state once block application is idle.
+      aborted = true;
+      finishStalled();
     }
   });
+
+  // Releases the sync state after a watchdog abort, but only once no block is
+  // mid-application. `applyBlock()` writes mem/account/round state outside
+  // `library.sequence`, so advancing the outer sequence while it runs could let
+  // a new sync or live block mutate the same state concurrently. We therefore
+  // wait for `modules.blocks.isActive` to clear before tearing down. If the
+  // in-flight series reaches a checkpoint first, it finishes the run itself and
+  // this becomes a no-op via the `settled` guard.
+  function finishStalled () {
+    if (settled) {
+      return;
+    }
+
+    if (modules.blocks.isActive.get()) {
+      library.logger.warn('loader', 'Sync watchdog: waiting for in-flight block application to finish before recovery', {
+        height: modules.blocks.lastBlock.get().height
+      });
+      return setTimeout(finishStalled, 1000);
+    }
+
+    // Terminal for this attempt: report no error so the async.retry wrapper does
+    // not immediately restart the sync; the sync timer starts a fresh attempt.
+    return finishSync(null);
+  }
 
   function finishSync (err) {
     if (settled) {
@@ -737,14 +769,14 @@ __private.sync = function (cb) {
     __private.blocksToSync = 0;
 
     // After an abort, drop the cached network so the next attempt re-selects peers.
-    if (__private.abortSync) {
+    if (aborted) {
       __private.initialize();
     }
 
     library.logger.info('loader', __private.stopRequested ? 'Sync stopped for shutdown' : 'Finished sync', {
       height: modules.blocks.lastBlock.get().height,
       stopped: __private.stopRequested,
-      aborted: __private.abortSync,
+      aborted: aborted,
       error: err ? err.message || err : null
     });
     library.bus.message('syncFinished');
@@ -752,7 +784,7 @@ __private.sync = function (cb) {
   }
 
   function skipOnStop (seriesCb, next) {
-    if (__private.isStopRequested()) {
+    if (shouldStop()) {
       return setImmediate(seriesCb);
     }
 
@@ -780,7 +812,9 @@ __private.sync = function (cb) {
       });
     },
     loadBlocksFromNetwork: function (seriesCb) {
-      return skipOnStop(seriesCb, __private.loadBlocksFromNetwork);
+      return skipOnStop(seriesCb, function (next) {
+        return __private.loadBlocksFromNetwork(next, shouldStop);
+      });
     },
     updateSystem: function (seriesCb) {
       return skipOnStop(seriesCb, function (next) {
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -2,6 +2,7 @@
 
 var async = require('async');
 var constants = require('../helpers/constants.js');
+var createSyncWatchdog = require('../helpers/syncWatchdog.js');
 var jobsQueue = require('../helpers/jobsQueue.js');
 var ip = require('neoip');
 var sandboxHelper = require('../helpers/sandbox.js');
@@ -25,6 +26,13 @@ __private.syncInterval = 10000;
 __private.retries = 5;
 __private.stopRequested = false;
 __private.shutdownRequested = false;
+// Set true when the watchdog aborts a stalled sync; signals the in-flight sync
+// to bail at its next safe checkpoint without disabling future syncs (unlike
+// `stopRequested`, which is reserved for shutdown).
+__private.abortSync = false;
+// Maximum time a single sync run may go without block progress before it is
+// considered stalled and aborted so the loader can recover.
+__private.syncTimeout = constants.syncStallTimeout;
 
 /**
  * Initializes library with scope content.
@@ -119,12 +127,14 @@ __private.requestStop = function () {
 };
 
 /**
- * Checks if shutdown was requested.
+ * Checks whether in-flight loader work should stop at its next safe boundary.
+ * True on shutdown (`stopRequested`) or when the watchdog aborted a stalled
+ * sync (`abortSync`).
  * @private
  * @return {boolean}
  */
 __private.isStopRequested = function () {
-  return __private.stopRequested;
+  return __private.stopRequested || __private.abortSync;
 };
 
 /**
@@ -605,7 +615,7 @@ __private.loadBlocksFromNetwork = function (cb) {
     } else {
       async.whilst(
           function (testCb) {
-            return testCb(null, !__private.stopRequested && !loaded && errorCount < 5);
+            return testCb(null, !__private.isStopRequested() && !loaded && errorCount < 5);
           },
           function (next) {
             var peer = network.peers[Math.floor(Math.random() * network.peers.length)];
@@ -687,16 +697,69 @@ __private.sync = function (cb) {
   });
   library.bus.message('syncStarted');
 
+  __private.abortSync = false;
   __private.isActive = true;
   __private.syncTrigger(true);
 
+  // Completion is funnelled through finishSync() and guarded by `settled` so the
+  // watchdog and the natural async.series completion are mutually exclusive: a
+  // stalled sync that the watchdog aborts releases the sync state exactly once,
+  // and a later (abandoned) completion of the series becomes a harmless no-op.
+  var settled = false;
+
+  var watchdog = createSyncWatchdog({
+    timeoutMs: __private.syncTimeout,
+    getHeight: function () {
+      return modules.blocks.lastBlock.get().height;
+    },
+    onStall: function (height) {
+      library.logger.error('loader', 'Sync watchdog: no block progress, aborting stalled sync to allow recovery', {
+        height: height,
+        blocksToSync: __private.blocksToSync,
+        timeoutMs: __private.syncTimeout
+      });
+      // Ask the in-flight series to bail at its next safe checkpoint, then
+      // release the sync state so the sync timer can start a fresh attempt.
+      __private.abortSync = true;
+      finishSync('Sync watchdog timeout: no block progress');
+    }
+  });
+
+  function finishSync (err) {
+    if (settled) {
+      return;
+    }
+    settled = true;
+    watchdog.stop();
+
+    __private.isActive = false;
+    __private.syncTrigger(false);
+    __private.blocksToSync = 0;
+
+    // After an abort, drop the cached network so the next attempt re-selects peers.
+    if (__private.abortSync) {
+      __private.initialize();
+    }
+
+    library.logger.info('loader', __private.stopRequested ? 'Sync stopped for shutdown' : 'Finished sync', {
+      height: modules.blocks.lastBlock.get().height,
+      stopped: __private.stopRequested,
+      aborted: __private.abortSync,
+      error: err ? err.message || err : null
+    });
+    library.bus.message('syncFinished');
+    return setImmediate(cb, err);
+  }
+
   function skipOnStop (seriesCb, next) {
-    if (__private.stopRequested) {
+    if (__private.isStopRequested()) {
       return setImmediate(seriesCb);
     }
 
     return next(seriesCb);
   }
+
+  watchdog.start();
 
   async.series({
     undoUnconfirmedList: function (seriesCb) {
@@ -742,17 +805,7 @@ __private.sync = function (cb) {
       });
     }
   }, function (err) {
-    __private.isActive = false;
-    __private.syncTrigger(false);
-    __private.blocksToSync = 0;
-
-    library.logger.info('loader', __private.stopRequested ? 'Sync stopped for shutdown' : 'Finished sync', {
-      height: modules.blocks.lastBlock.get().height,
-      stopped: __private.stopRequested,
-      error: err ? err.message || err : null
-    });
-    library.bus.message('syncFinished');
-    return setImmediate(cb, err);
+    return finishSync(err);
   });
 };
 
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -708,11 +708,28 @@ __private.sync = function (cb) {
   // sync state is released exactly once, and a later (abandoned) completion of
   // the series becomes a harmless no-op.
   var settled = false;
+  // True while a sync phase that mutates unconfirmed account/pool state is in
+  // flight. Block application is tracked separately by `modules.blocks.isActive`.
+  // The watchdog must not tear the run down while either is set, otherwise it
+  // could advance `library.sequence` while an abandoned mutation is still alive.
+  var mutatingState = false;
 
   // Shutdown or this run's watchdog abort both stop the in-flight series at its
   // next safe checkpoint.
   function shouldStop () {
     return __private.stopRequested || aborted;
+  }
+
+  // Wraps a state-mutating sync phase so the watchdog defers teardown until the
+  // phase's callback has actually returned.
+  function mutating (run) {
+    return function (next) {
+      mutatingState = true;
+      return run(function (err) {
+        mutatingState = false;
+        return next(err);
+      });
+    };
   }
 
   var watchdog = createSyncWatchdog({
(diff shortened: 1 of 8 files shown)


modules/loader.js
```diff
@@ -687,17 +697,70 @@ __private.sync = function (cb) {
...
+    onStall: function (height) {
+      library.logger.error('loader', 'Sync watchdog: no block progress, aborting stalled sync to allow recovery', {
+        height: height,
+        blocksToSync: __private.blocksToSync,
+        timeoutMs: __private.syncTimeout
+      });
+      // Ask the in-flight series to bail at its next safe checkpoint, then
+      // release the sync state so the sync timer can start a fresh attempt.
+      __private.abortSync = true;
+      finishSync('Sync watchdog timeout: no block progress');
```
al-onyxprotocol: Blocking: this timeout is returned as an error to the `async.retry(__private.retries, __private.sync, sequenceCb)` wrapper in `syncTimer`. With async@3.2.6 that immediately starts the next retry in the same `library.sequence` task, so the recovery does not wait for the next timer tick. The new retry executes line 700 and clears the global `abortSync` flag while the previous hung/slow flow may still be alive, so the abandoned flow can lose its stop signal and keep mutating state. Please make watchdog abort terminal for the current attempt without feeding retry this way, or isolate the abort signal per run and prove old work cannot resume.

modules/loader.js
```diff
@@ -687,17 +697,70 @@ __private.sync = function (cb) {
...
+  });
+
+  function finishSync (err) {
+    if (settled) {
+      return;
+    }
+    settled = true;
+    watchdog.stop();
+
+    __private.isActive = false;
```
al-onyxprotocol, line 792: Blocking: `finishSync()` releases `loader.syncing()` and advances the outer sequence before the in-flight `async.series` necessarily reaches a safe checkpoint. If the watchdog fires while `modules.blocks.verify.processBlock()` / `modules.blocks.chain.applyBlock()` is still running, `loadBlocksFromPeer()` only checks `shouldStop` before starting the next block, not inside the current block. Since `applyBlock()` performs mem/account/round writes outside `library.sequence`, the next sync or live block can start before the old block application finishes. That breaks the sequencing invariant and can corrupt consensus-derived memory state. The watchdog should not advance the outer sequence until current state mutation is known to be idle, or live/new sync admission must remain blocked while `modules.blocks.isActive` / the old run is active.

al-onyxprotocol:
Review summary

What looks good:
- The issue is valid: a stuck callback in the sync flow can leave `loader.syncing()` true forever, reject live blocks, and prevent recovery sync attempts.
- The PR is aimed at liveness/recovery and does not directly change block serialization, IDs, signatures, rewards, fees, activation heights, or validation rules.
- The new watchdog helper is small, isolated, and has focused unit coverage.

What needs attention before merge:
- I do not think this is merge-ready yet because the abort path can break the sequencing guarantees that normally protect block/state mutation.
- The watchdog currently returns an error into `async.retry()`, which immediately starts new sync attempts in the same sequence task. That can also clear the global `abortSync` flag while an abandoned previous flow is still alive.
- `finishSync()` releases `loader.syncing()` / the outer sequence before proving that the old in-flight block processing has reached a safe checkpoint. If the watchdog fires during `processBlock()` / `applyBlock()`, new sync or live-block processing can begin while old mem/account/round writes are still running.

Findings:
- Blocking: watchdog timeout is fed into `async.retry`, causing immediate retries and potentially clearing the previous run's abort signal. See the inline comment on `modules/loader.js`.
- Blocking: the abort path can advance the outer sequence before current block state mutation is idle, which risks concurrent consensus-derived state writes. See the inline comment on `modules/loader.js`.

Security / consensus assessment:
- I did not find a direct weakening of signature checks, peer handshake checks, block validation, or serialization.
- The main risk introduced by this PR is concurrency/data-integrity related. In this node, that is consensus-sensitive because `mem_*`, rounds, balances, and `lastBlock` state must remain sequenced.

Validation run:
- `npm run test:single -- test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed: 14 passing.
- `npx eslint helpers/syncWatchdog.js modules/loader.js helpers/constants.js test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed.
- `git diff --check origin/dev...HEAD` passed.
- `npm run test:unit:fast` was attempted but blocked by the local native `libpq` module being built for NODE_MODULE_VERSION 127 while the current Node.js runtime requires 147; this appears to be local environment drift, not a PR test failure.

modules/loader.js
massivedev0: Fixed — both halves of this finding.

**No more retry storm.** The watchdog no longer feeds an error into `async.retry`. On stall it routes through `finishStalled()` → `finishSync(null)`, i.e. the attempt is reported as terminal success, so `async.retry` does **not** immediately restart it. Recovery now waits for the next `syncTimer` tick instead of re-entering in the same sequence task.

**Abort signal is now run-scoped.** The global `__private.abortSync` is gone. Each `__private.sync` run owns a local `var aborted` and exposes a per-run `shouldStop()` that closes over it; that predicate is threaded into `__private.loadBlocksFromNetwork(next, shouldStop)` and on into `loadBlocksFromPeer(..., shouldStop)`. So a fresh run gets its own flag and can never clear an earlier run's signal, and an abandoned run keeps `aborted === true` forever. Because `loadBlocksFromPeer` checks `shouldStop` before each block, a resumed-but-abandoned run bails before it can start another `applyBlock`. This is the "isolate the abort signal per run and prove old work cannot resume" option you offered.

modules/loader.js
massivedev0: Fixed — the outer sequence is no longer advanced while a block is mid-application.

The watchdog no longer calls `finishSync()` directly. It now calls `finishStalled()`, which checks `modules.blocks.isActive.get()`: while a block is being applied it logs and re-checks every 1s, and only releases `loader.syncing()` / advances the sequence once `isActive` is clear. You're right that `applyBlock()` (`modules/blocks/chain.js:339-503`) does its `mem_*`/account/round writes outside `library.sequence` and only guards them with `modules.blocks.isActive` — that's exactly the signal `finishStalled()` now waits on, so a new sync or live block can't begin mutation while the old write is in flight.

Combined with the run-scoped `shouldStop()` (see the other thread), `loadBlocksFromPeer` checks the stop predicate before each block, so an aborted run cannot start a *new* `applyBlock` either. The two together close the concurrent-mutation window.

Deliberate limit: if `isActive` never clears (a DB write genuinely hung mid-application — distinct from the observed incident, where Postgres was idle and the flow was parked in a callback that never fired), `finishStalled()` keeps waiting and logging rather than forcing teardown, because forcing it is exactly what could corrupt consensus-derived memory tables. That catastrophic case needs operator intervention and is out of scope for this liveness fix.

massivedev0:
Thanks for the careful review — both blocking findings were valid and are addressed. Summary of changes (replies on the two threads have specifics):

1. **Watchdog abort is no longer fed to `async.retry`** → no immediate in-sequence retry; recovery waits for the next `syncTimer` tick.
2. **Abort signal is run-scoped** (`var aborted` + per-run `shouldStop()` threaded through `loadBlocksFromNetwork`/`loadBlocksFromPeer`), so a fresh run can't clear an older run's stop signal and an abandoned run can't start another `applyBlock`.
3. **Teardown is gated on `modules.blocks.isActive`** via a new `finishStalled()`: `loader.syncing()` is released only when no block is mid-application, so the outer sequence is never advanced while `applyBlock()` is writing `mem_*`/round/account state.

Net: liveness-only recovery that cannot advance the sequence into a concurrent state mutation. A slow-but-progressing sync is still never aborted (height-progress check).

Validation: `test:single` for `syncWatchdog` + `blocks.process` (14 passing) and `test:unit:fast` (855 passing) locally; eslint clean on touched files.

Note on probable root cause (out of scope for this safety-net PR): `modules/blocks/verify.js` `checkExists` does `library.db.query(...).then(...)` with **no `.catch`** — a rejected query there would never call `seriesCb`, hanging `processBlock` *before* `applyBlock` (so `isActive` is false, which this watchdog recovers cleanly). Worth a focused follow-up to add the missing rejection handler.

Heads-up: these changes are staged locally for author review and not yet pushed, so the updated diff isn't visible on the PR yet — it'll appear once pushed.

modules/loader.js
```diff
@@ -690,14 +699,100 @@ __private.sync = function (cb) {
...
+  // a new sync or live block mutate the same state concurrently. We therefore
+  // wait for `modules.blocks.isActive` to clear before tearing down. If the
+  // in-flight series reaches a checkpoint first, it finishes the run itself and
+  // this becomes a no-op via the `settled` guard.
+  function finishStalled () {
+    if (settled) {
+      return;
+    }
+
+    if (modules.blocks.isActive.get()) {
```
massivedev0: Blocking: this guard only waits for `modules.blocks.isActive`, but the watchdog is started before the sync run enters `undoUnconfirmedList`, and the run later also calls `applyUnconfirmedList`. Those paths mutate unconfirmed account/pool state (`logic/transactionPool.js` calls `modules.transactions.undoUnconfirmed()` / `applyUnconfirmed()`), but they do not set `blocks.isActive` unless they are reached through `applyBlock()`. If the watchdog fires while the loader's standalone `undoUnconfirmedList` or `applyUnconfirmedList` is slow or hung, `finishStalled()` will release `loader.syncing()` and advance `library.sequence` while that old state mutation may still be alive, allowing a fresh sync or live block to run the same unconfirmed-state transitions concurrently. Please either limit the watchdog teardown to phases where no state mutation is in flight, or track/wait for all sync-run state-mutating phases, not only block application.

massivedev0:
Re-review summary after `55eacf5382b6403010ce9076c995176117658e32`

What improved:
- The previous `async.retry` concern is addressed: watchdog abort now completes the attempt with `null`, so it does not immediately trigger retry inside the same `library.sequence` task.
- The previous global `abortSync` concern is addressed: abort state is now run-scoped, so a new run cannot clear an older run's stop predicate.
- Waiting for `modules.blocks.isActive` before releasing the sync state correctly covers the most dangerous `applyBlock()` / mem-round-account write window.

Remaining finding:
- Blocking: the watchdog still starts before `undoUnconfirmedList` and the sync flow later calls `applyUnconfirmedList`; both mutate unconfirmed account/pool state without setting `modules.blocks.isActive` unless reached through `applyBlock()`. If the watchdog fires during one of these standalone loader phases, it can still release `loader.syncing()` and advance `library.sequence` while an old state-mutating callback is alive. I left an inline comment on `modules/loader.js`.

Security / consensus assessment:
- I still do not see direct changes to block serialization, IDs, signatures, reward/fee logic, delegate ordering, activation gates, nethash checks, or public API auth/validation.
- The remaining risk is concurrency/data-integrity, not an intentional consensus-rule change; however, in this codebase concurrent mem/account/unconfirmed state transitions are regression-prone and can affect block/transaction processing safety.

Validation run:
- `npm run test:single -- test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed: 14 passing.
- `npx eslint helpers/syncWatchdog.js modules/loader.js helpers/constants.js test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed.
- `git diff --check origin/dev...HEAD` passed.
- `npm run test:unit:fast` was attempted again but remains blocked by the local native `libpq` module being built for NODE_MODULE_VERSION 127 while this Node.js runtime requires 147.

modules/loader.js
massivedev0: Fixed — teardown now waits for the unconfirmed-state phases too, not only `applyBlock()`.

Confirmed the gap: `undoUnconfirmedList`/`applyUnconfirmedList` go through `logic/transactionPool.js` with **no `balancesSequence`**, so they were serialized only by the outer `library.sequence` — exactly the guarantee the watchdog could break.

Change: a run-local `mutatingState` flag is set around both unconfirmed phases (via a small `mutating()` wrapper), and `finishStalled()` now defers teardown while `mutatingState || modules.blocks.isActive.get()`. So the run is released only when **no** sync-phase state mutation is in flight — block application *or* unconfirmed undo/apply.

I deliberately did **not** mark the whole `loadBlocksFromNetwork` phase as mutating: its only state mutation is `applyBlock()`, already covered precisely by `modules.blocks.isActive`. Wrapping the entire phase would block recovery during the benign parts (peer fetch / `getCommonBlock` / `checkExists` waits) — which is where the observed stall actually parks (`isActive` false, Postgres idle). Keeping the gate per-mutation preserves recovery there while still never advancing `library.sequence` mid-write.

Same deliberate limit as before: if a mutating phase is genuinely hung (never returns), `finishStalled()` keeps waiting and logging rather than forcing teardown into a possible concurrent mutation; that catastrophic case needs operator intervention and is out of scope for a liveness fix.

massivedev0:
Re-review finding addressed (commit pending push).

The standalone unconfirmed phases (`undoUnconfirmedList` / `applyUnconfirmedList`) are confirmed to run without `balancesSequence`, so the watchdog could previously advance `library.sequence` while one of them was still mutating unconfirmed account/pool state. Fix:

- Added a run-local `mutatingState` flag, set around both unconfirmed phases.
- `finishStalled()` now waits on `mutatingState || modules.blocks.isActive.get()` before releasing `loader.syncing()` / advancing the sequence.
- `applyBlock()` mutation stays covered by `modules.blocks.isActive`; the non-mutating network parts of `loadBlocksFromNetwork` remain interruptible so the common stall (parked peer/DB callback, `isActive` false) still recovers.

So teardown now happens only when no sync-phase state mutation — block apply or unconfirmed undo/apply — is in flight.

Validation: `test:single` (`syncWatchdog` + `blocks.process`) 14 passing; `test:unit:fast` 855 passing; eslint clean on `modules/loader.js`.

Note: this is staged locally for author review and not yet pushed, so the diff isn't visible on the PR yet.

modules/loader.js
```diff
@@ -690,22 +699,129 @@ __private.sync = function (cb) {
...
+  // therefore wait for both `modules.blocks.isActive` and the run-local
+  // `mutatingState` to clear before tearing down. If the in-flight series reaches
+  // a checkpoint first, it finishes the run itself and this becomes a no-op via
+  // the `settled` guard.
+  function finishStalled () {
+    if (settled) {
+      return;
+    }
+
+    if (mutatingState || modules.blocks.isActive.get()) {
```
al-onyxprotocol, line 771: Blocking: this still does not cover an already-started `loadBlocksFromPeer()` block processing run before it enters `applyBlock()`. `loadBlocksFromPeer()` checks `shouldStop` only before calling `modules.blocks.verify.processBlock()`. Inside `processBlock()`, the async `checkExists`, `validateBlockSlot`, and `checkTransactions` steps can take longer than `syncStallTimeout`; during those steps `mutatingState` is false and `modules.blocks.isActive` is still false because `applyBlock()` has not started yet. If the watchdog releases `loader.syncing()` in that window, a fresh sync/live block can start; then the old `processBlock()` may later resume and enter `applyBlock()`, creating exactly the concurrent block/state mutation this guard is trying to prevent. Please track the in-flight block-processing phase, or thread a stop check into `processBlock()` before it can transition into `applyBlock()` after an abort.

al-onyxprotocol:
Re-review summary after `3f23ce461747116a574c7b757a6eff46b68a72d9`

What improved:
- The previous standalone unconfirmed-state concern is addressed: `undoUnconfirmedList` and `applyUnconfirmedList` are now wrapped with run-local `mutatingState`, and `finishStalled()` waits for that flag before releasing the sync state.
- The earlier `async.retry` and global abort-signal concerns remain addressed.
- I still do not see direct changes to serialization, signatures, block IDs, rewards, fees, delegate ordering, activation gates, nethash checks, or API auth/validation.

Remaining finding:
- Blocking: there is still an uncovered window after `loadBlocksFromPeer()` starts `modules.blocks.verify.processBlock()` but before `applyBlock()` sets `modules.blocks.isActive`. If the watchdog fires while `processBlock()` is in `checkExists`, `validateBlockSlot`, or `checkTransactions`, `finishStalled()` can release `loader.syncing()` because both `mutatingState` and `blocks.isActive` are false; the old `processBlock()` can later resume and enter `applyBlock()` concurrently with a new sync/live block. I left an inline comment on `modules/loader.js`.

Validation run:
- `npm run test:single -- test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed: 14 passing.
- `npx eslint helpers/syncWatchdog.js modules/loader.js helpers/constants.js test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js` passed.
- `git diff --check origin/dev...HEAD` passed.
- `npm run test:unit:fast` was attempted again but remains blocked by local native dependency drift: `libpq` was built for NODE_MODULE_VERSION 127 while this Node.js runtime requires 147.

modules/loader.js
massivedev0: You're right, and thanks for pushing on this — the per-phase flag chase was the wrong shape. Reworked it into a single invariant instead of another window patch.

**Invariant: once a run is aborted, it can never *begin* a new state mutation; teardown only waits for a mutation that had already started.**

The window you found (parked in `checkExists`/`validateBlockSlot`/`checkTransactions`, both flags false) is closed at the source: the run's `shouldStop()` is now threaded `loadBlocksFromPeer → verify.processBlock`, and `processBlock` checks it **right before `applyBlock()`** — the actual chain-state mutation. So no matter where a block was parked when the watchdog fired, when it resumes it verifies but bails before applying. It can't enter `applyBlock()` concurrently with a fresh run.

With new applies blocked at that boundary (and new unconfirmed phases blocked by `skipOnStop`), releasing `loader.syncing()` is safe even while an abandoned flow is still parked. `finishStalled()` then only has to drain a mutation that was *already in progress* at abort — `modules.blocks.isActive` (a block apply mid-write) and `mutatingState` (an unconfirmed undo/apply mid-write) — both bounded, ms-scale.

So the recovery no longer depends on enumerating every pre-apply window; it depends on the single fact that an aborted run cannot apply.

Diff: gate in `modules/blocks/verify.js` (`processBlock` gains an optional `shouldStop`, checked before `applyBlock`); pass-through in `modules/blocks/process.js` `loadBlocksFromPeer`; `modules/loader.js` threads the run-scoped `shouldStop`. New unit tests in `test/unit/modules/blocks.verify.js` assert a block is verified but not applied when `shouldStop()` is true (and applied normally otherwise / when no predicate is given).

One honest residual: the fork-2 path inside `checkTransaction` calls `undoUnconfirmed` *before* the apply gate. It only fires on an already-confirmed tx (a real fork) for a block mid-verification at the exact abort instant, and it mutates only the volatile unconfirmed pool (rebuilt, not persisted consensus state). I judged a check there to be a band-aid for a practically-impossible, non-persistent case rather than part of the invariant — happy to add a guard if you'd prefer belt-and-suspenders. Note the underlying hang itself goes away once the missing `.catch` in `checkExists` is added (separate follow-up).

massivedev0:
Re-review finding addressed — and reworked into one coherent mechanism rather than another per-window flag (commit pending push).

**Root of the back-and-forth:** I was trying to *detect* "is a mutation in flight" for each sync phase, which is inherently incomplete (you correctly found the pre-`applyBlock` window each time). Replaced that with a single invariant:

> Once a sync run is aborted, it can never *begin* a new state mutation. Teardown only waits for a mutation already in progress to drain.

**How it's enforced now:**
- **Block apply** — the run's `shouldStop()` is threaded `loadBlocksFromPeer → verify.processBlock` and checked **immediately before `applyBlock()`**. A block parked anywhere in verification (`checkExists`/`validateBlockSlot`/`checkTransactions`) bails before applying when it resumes. This closes all pre-apply windows at the boundary instead of enumerating them.
- **Unconfirmed undo/apply phases** — not started after abort (`skipOnStop`).
- **Drain** — `finishStalled()` releases `loader.syncing()` only once `modules.blocks.isActive` (in-flight block apply) and `mutatingState` (in-flight unconfirmed phase) are clear. Those are bounded, already-started writes.

So recovery no longer depends on enumerating mutation windows; it rests on "an aborted run cannot apply."

**Touched:** `modules/blocks/verify.js` (optional `shouldStop` on `processBlock`, gate before `applyBlock`), `modules/blocks/process.js` (`loadBlocksFromPeer` pass-through), `modules/loader.js` (thread run-scoped `shouldStop`; `finishStalled` drains in-flight only). For non-sync callers (`onReceiveBlock`, `generateBlock`) `shouldStop` is absent → zero behavior change.

**Residual (documented, not patched):** the fork-2 `undoUnconfirmed` inside `checkTransaction` runs before the apply gate, but only on an already-confirmed tx for a block mid-verification at the exact abort instant, and mutates only the volatile/rebuilt unconfirmed pool — not persisted consensus state. Can add a guard if you want it explicit. The hang that makes the watchdog fire is itself removed by adding the missing `.catch` in `checkExists` (separate follow-up).

**Validation:** `test:single` for `blocks.verify` (new, 3 passing), `blocks.process` and `syncWatchdog`; `test:unit:fast` → 858 passing; eslint clean on all touched files.

Note: staged locally for author review, not yet pushed — diff will appear once pushed.

al-onyxprotocol:
Re-review summary after `ea1fdd8ab9fbf1d9da2a7dde5132ac94a4f0accf`

What looks good:
- The task remains valid: a hung callback in the sync path can leave `loader.syncing()` true, reject live blocks, and prevent automatic recovery even when the local chain is otherwise valid.
- The latest fix addresses the previous blocking race. The run-scoped `shouldStop()` predicate is now threaded through `loadBlocksFromNetwork -> loadBlocksFromPeer -> verify.processBlock`, and `processBlock()` checks it immediately before `applyBlock()`. If the watchdog aborts while verification is parked in `checkExists`, slot validation, or transaction checks, the old run now returns `Sync aborted before applying block` instead of starting block application.
- The aborted block is not reported as successfully loaded: `loadBlocksFromPeer()` receives an error, does not update `lastValidBlock`, and the loader's late completion is guarded by `settled`.
- Non-sync callers (`generateBlock` / live block receive) still call `verify.processBlock()` without `shouldStop`, so their behavior is unchanged.
- I do not see changes to block/transaction serialization, IDs, signatures, rewards, fees, delegate ranking, activation heights, nethash checks, or public API validation.

What to keep an eye on:
- There is still a pre-apply fork-2 path in `verify.checkTransaction()` that can call `transactions.undoUnconfirmed()` for an already-confirmed transaction before the new `shouldStop` gate. I do not consider it a merge blocker here because it is limited to volatile unconfirmed account/pool state on an invalid/fork block, not persisted consensus state, but it is worth keeping as a follow-up if we want the abort boundary to be completely explicit for every pre-apply state write.
- The probable root cause mentioned earlier (`checkExists` missing a `.catch`) is still a good separate follow-up. This PR is a liveness safety net rather than the root-cause fix.
- The PR body's testing section still lists the older targeted command without `test/unit/modules/blocks.verify.js`; not a code issue, but it may be worth updating for reviewer clarity.

Findings:
- No new blocking findings after the latest fix.
- The previous `async.retry`, global abort signal, standalone unconfirmed-phase, and pre-`applyBlock()` race concerns are addressed.

Validation run locally:
- `node -v` -> `v26.0.0`
- `npm run test:single -- test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js test/unit/modules/blocks.verify.js` -> passed, 17 passing.
- `npx eslint helpers/syncWatchdog.js modules/loader.js helpers/constants.js modules/blocks/process.js modules/blocks/verify.js test/unit/helpers/syncWatchdog.js test/unit/modules/blocks.process.js test/unit/modules/blocks.verify.js` -> passed.
- `git diff --check origin/dev...origin/pr/249` -> passed.
- `npm run test:unit:fast` -> attempted, but blocked by local native dependency drift: `node_modules/libpq/build/Release/addon.node` was compiled for `NODE_MODULE_VERSION 127`, while this runtime requires `NODE_MODULE_VERSION 147`. The suite starts and the touched focused tests pass before DB-backed hooks fail on that ABI mismatch.
- `gh pr checks 249 --repo Adamant-im/adamant` -> no checks reported on the branch.

massivedev0:
Thanks for the clear re-review. Addressed the "keep an eye on" items (commit pending push):

**1. Pre-apply fork-2 write — now gated.** `checkTransaction()` takes the run's `shouldStop`, and when the run is aborted it skips the `transactions.undoUnconfirmed()` / `removeUnconfirmedTransaction()` writes (the block is rejected anyway and the unconfirmed pool is rebuilt). With this, the abort boundary is explicit for **every** pre-apply state write, not just `applyBlock()`. `delegates.fork(block, 2)` is left in place — it records a diagnostic `forks_stat` row, not consensus/mem state.

**2. `checkExists` missing `.catch` — kept as a separate follow-up**, as you suggested. This PR stays the liveness safety net; the root-cause rejection handler is tracked separately.

**3. PR body testing section — updated** to include `test/unit/modules/blocks.verify.js`, the corrected `test:unit:fast` count (861), the full eslint file list, and a per-suite coverage breakdown. I also refreshed the Description/Fix and Risk sections to describe the final `shouldStop`-gate design rather than the earlier `abortSync` wording.

**Test coverage (all cases):**
- `test/unit/helpers/syncWatchdog.js` — stall detection (fires on no progress, ignores progressing sync, fires once, disabled at `timeoutMs <= 0`, `stop()`, idempotent `start()`).
- `test/unit/modules/blocks.verify.js` — apply-boundary gate (verify-but-not-apply when aborted; applies otherwise / when no predicate); **fork-2 gate** (undo skipped when aborted, performed otherwise).
- `test/unit/modules/blocks.process.js` — `isReadyToReceiveBlock()` branches + `loadBlocksFromPeer` forwards `shouldStop` to `verify.processBlock`.

Validation: `test:single` for the three suites passes; `test:unit:fast` → **861 passing**; eslint clean on all touched files.

Note: staged locally for author review, not yet pushed — the diff (fork-2 gate + new tests) will appear once pushed.

al-onyxprotocol (approved):
👍 it's good now

PR #251, opened 2026-06-29
Fix: Handle DB query rejections in block/sync path to prevent silent stalls

### LINKED ISSUE #250: [Bug] Unhandled DB query rejections in block/sync path can silently stall processing
## Summary

Several `library.db.*(...).then(...)` chains in the block/sync path have **no `.catch`**. If the query rejects (DB hiccup, disconnect, timeout), the `.then` never runs and the callback the chain owns is never invoked — silently stalling block processing. In the sync loader this is a probable root cause of the "stuck sync" wedge addressed by the liveness watchdog in #248 / #249 (that PR recovers from such a stall; this one removes a source of it).

## Affected sites

- `modules/blocks/verify.js` — `processBlock()` → `checkExists` step: `db.query(getBlockId).then(...)` with no `.catch`. A rejection hangs `processBlock` before `applyBlock`, i.e. exactly the parked-callback signature observed on the stalled node (Postgres idle, event loop idle).
- `modules/blocks/utils.js` — `loadBlocksData()`: the **inner** `db.query(loadBlocksData).then(...)` lacks a `.catch` (the outer `getHeightByLastId` query already has one). A rejection stalls the `dbSequence` task that owns it.
- `modules/loader.js` — `loadBlockChain()`: the inner `db.task(updateMemAccounts)` promise is not returned and has no `.catch`, so its rejection is unhandled and blockchain loading can stall at startup.

Other `db.*` chains in `modules/blocks/process.js`, `chain.js`, and `api.js` were scanned and already have rejection handlers.

## Fix

Add a `.catch` that surfaces the error through the owning callback (`setImmediate(seriesCb/cb, err)`), and for the loader case `return` the inner `db.task` so the rejection propagates to the existing outer `.catch`. This turns a silent hang into a normal error, so the sync simply retries another peer / a normal error path runs instead of wedging.

## Scope / risk

- Pure error-handling; no change to the success path, block validation, serialization, IDs, signatures, rewards, rounds, or activation gating.
- Liveness/robustness only. Complements the watchdog (#248/#249): watchdog = recover from any stall; this = remove a concrete source of stalls.

Related: #248, #249

massivedev0:
## Description

Root-cause companion to the sync-stall watchdog (#249). Several `library.db.*(...).then(...)` chains in the block/sync path have **no `.catch`**, so a rejected query (DB hiccup, disconnect, timeout) never runs the `.then` and never invokes the callback the chain owns — silently stalling block processing. That parked-callback behavior (Postgres idle, event loop idle) is the probable trigger of the stuck-sync wedge the watchdog recovers from. This PR removes the source; the watchdog stays the safety net.

### Fixes

- **`modules/blocks/verify.js`** — `processBlock()` `checkExists`: add `.catch` → `Blocks#checkExists error`. A rejection here parked `processBlock` *before* `applyBlock`, matching the observed stall signature.
- **`modules/blocks/utils.js`** — `loadBlocksData()`: add `.catch` to the **inner** `loadBlocksData` query (the outer `getHeightByLastId` query already had one) so a rejection ends the `dbSequence` task instead of hanging it.
- **`modules/loader.js`** — `loadBlockChain()`: `return` the inner `db.task(updateMemAccounts)` so its rejection propagates to the existing outer `.catch` instead of being swallowed (could stall startup).

Scanned the rest of the block subsystem (`process.js`, `chain.js`, `api.js`) — those `db.*` chains already have rejection handlers.

## Related issue

Closes #250 (companion to #248 / #249)

## How to test

```
npm run test:single -- test/unit/modules/blocks.verify.js test/unit/modules/blocks.utils.js
```

- `blocks.verify.js` — a rejected `checkExists` query fails the series with `Blocks#checkExists error` and does not call `applyBlock` (instead of hanging).
- `blocks.utils.js` — a rejected inner `loadBlocksData` query calls back with `Blocks#loadBlockData error` (instead of stalling the `dbSequence` task).

Regression: `npm run test:unit:fast` → 863 passing. Lint clean on all touched files (`npx eslint modules/blocks/verify.js modules/blocks/utils.js modules/loader.js test/unit/modules/blocks.verify.js test/unit/modules/blocks.utils.js`).

Not run: `npm run test:api` (needs a running testnet).

## Risk areas

- **Consensus / protocol:** none — pure error handling; success paths, block validation, serialization, IDs, signatures, rewards, rounds, and activation gating are untouched.
- **Behavior change:** a DB query rejection that previously hung now surfaces as a normal error, so the loader retries another peer / runs its existing error path. For `loadBlockChain` the rejection now reaches the existing outer `.catch` (logs and triggers cleanup) instead of hanging startup.

## Checklist

- [x] No security checks weakened.
- [x] Schema/SQL unaffected.
- [x] Tests added and run (targeted + `test:unit:fast`).

### DIFF
--- a/modules/blocks/utils.js
+++ b/modules/blocks/utils.js
@@ -277,6 +277,11 @@ Utils.prototype.loadBlocksData = function (filter, options, cb) {
       // FIXME: That SQL query have mess logic, need to be refactored
       library.db.query(sql.loadBlocksData(filter), params).then(function (rows) {
         return setImmediate(cb, null, rows);
+      }).catch(function (err) {
+        // A rejection here would otherwise never call back, stalling the
+        // dbSequence task that owns this query.
+        library.logger.error('blocks', `Failed to load blocks data: ${err?.message || err}`, err && err.stack);
+        return setImmediate(cb, 'Blocks#loadBlockData error');
       });
     }).catch(function (err ) {
       library.logger.error('blocks', `Failed to get height by last id ${filter.lastid}: ${err?.message || err}`, err.stack);
--- a/modules/blocks/verify.js
+++ b/modules/blocks/verify.js
@@ -461,6 +461,11 @@ Verify.prototype.processBlock = function (block, broadcast, cb, saveBlock, shoul
         } else {
           return setImmediate(seriesCb);
         }
+      }).catch(function (err) {
+        // Without this handler a rejected query never resolves the series step,
+        // which silently stalls block processing (and can wedge the sync loader).
+        library.logger.error('blocks', `Failed to check block existence for ${block.id}: ${err?.message || err}`, err && err.stack);
+        return setImmediate(seriesCb, 'Blocks#checkExists error');
       });
     },
     validateBlockSlot: function (seriesCb) {
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -558,7 +558,9 @@ __private.loadBlockChain = function () {
       return t.batch(promises);
     }
 
-    library.db.task(updateMemAccounts).then(function (results) {
+    // Returned so a rejection propagates to the outer .catch below instead of
+    // being swallowed (which would silently stall blockchain loading).
+    return library.db.task(updateMemAccounts).then(function (results) {
       if (__private.stopRequested) {
         return finishForShutdown();
       }
--- a/test/unit/modules/blocks.utils.js
+++ b/test/unit/modules/blocks.utils.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const { expect } = require('chai');
+const sinon = require('sinon');
+
+const Utils = require('../../../modules/blocks/utils.js');
+
+// Focused test for the missing `.catch` on the inner loadBlocksData query: a
+// rejected query must call back with an error instead of stalling the
+// dbSequence task that owns it.
+describe('blocks utils - loadBlocksData query rejection', function () {
+  let utils;
+  let logger;
+  let db;
+  let dbSequence;
+
+  beforeEach(function () {
+    logger = { trace: sinon.spy(), debug: sinon.spy(), error: sinon.spy(), info: sinon.spy() };
+    db = { query: sinon.stub() };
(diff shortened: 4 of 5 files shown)


massivedev0:
## Review summary

What looks good:

- The task is valid: the three changed sites could previously swallow rejected DB promises and leave the owning callback/sequence unresolved.
- The PR addresses the reported failure mode directly: `processBlock()` and `loadBlocksData()` now callback on DB rejection, and `loadBlockChain()` now returns the inner `db.task()` so the existing outer `.catch` handles failures.
- I did not find consensus-impacting changes. The success paths for block validation/application, serialization, IDs, signatures, rewards, rounds, SQL shape, and activation-gated behavior are unchanged.
- I did not find a new security weakness. The change tightens operational failure handling and does not bypass validation, signature checks, peer compatibility checks, or sequencing guards.

What to watch:

- `modules/loader.js` does not have a dedicated regression test for the returned inner `db.task()` rejection path. The implementation is straightforward and uses the existing outer cleanup path, so I do not consider this blocking, but it is the remaining test gap.
- The PR body's literal `npx eslint ...` command was not reproducible in my checkout because `npx` resolved ESLint 9 without finding the legacy config from the file paths. The repository lint task (`npm run eslint`) passed, so this looks like a command/tooling note rather than a code issue.
- GitHub reports no checks for this branch at the moment.

Findings:

- No blocking findings.
- No inline code comments left; I did not find a significant line-level issue worth commenting.

Verification run:

- `npm run test:single -- test/unit/modules/blocks.verify.js test/unit/modules/blocks.utils.js` -> 7 passing.
- `npm run eslint -- --gruntfile Gruntfile.js` -> passed.
- `git diff --check origin/dev...HEAD` -> passed.
- `npm run test:unit:fast` -> attempted, but blocked by local native dependency drift: `node_modules/libpq/build/Release/addon.node` was compiled for `NODE_MODULE_VERSION 127`, while this runtime requires `NODE_MODULE_VERSION 147`. The touched focused tests pass before the DB-backed suites fail on that ABI mismatch.

al-onyxprotocol (approved):
Nice

PR #253, opened 2026-06-30
Fix: Compute delegates/get rank and rate over the full delegate list (regression)

### LINKED ISSUE #252: [Bug] delegates/get returns rank/rate = 1 for every delegate (regression since v0.9.0)
### Summary

`GET /api/delegates/get` returns `rank: 1` and `rate: 1` for **every** delegate, regardless of the delegate's real position in the sorted delegate list. This is a regression: the same endpoint on an older node version returns the correct rank/rate. Because `rank`/`rate` are derived from the full sorted delegate set, any single-delegate lookup now reports the wrong value.

### Steps to reproduce

1. Query a single delegate on a node running an older release (correct behavior):
   - https://bid.adamant.im/api/delegates/get?username=kabadas
   - Response contains: `"rate":82,"rank":82`
2. Query the same delegate on a node running the current release (buggy behavior):
   - https://endless.adamant.im/api/delegates/get?username=kabadas
   - Response contains: `"rate":1,"rank":1`
3. Repeat for any other `username`/`publicKey`/`address` — the new node always returns `rank: 1` and `rate: 1`.

### Expected behavior

`rank` (and the deprecated alias `rate`) must reflect the delegate's real position in the full delegate ranking, computed over **all** delegates ordered by `votesWeight`/`vote` + `publicKey`. For `kabadas` that is `rank: 82`, `rate: 82`.

### Actual behavior

`rank` and `rate` are always `1` for any single-delegate lookup via `delegates/get`.

### Root cause

The regression was introduced in commit `60d7733c` (`feat: find single delegate by the given address`, first released in `v0.9.0`). It pushed the per-delegate filter down into the SQL/accounts query, so `rank`/`rate` are computed over a single-row result instead of the full list.

- `getDelegate` (`modules/delegates.js:784`) builds a `filter` (`address`/`username`) for the requested delegate and passes it to `getDelegates(req.body, filter, cb)` (`modules/delegates.js:808`).
- `getDelegates` (`modules/delegates.js:442`) spreads that `filter` straight into the accounts query:

  ```js
  modules.accounts.getAccounts({
    ...filter,          // e.g. { username: 'kabadas' }
    isDelegate: 1,
    sort: sortFilter
  }, [...], function (err, delegates) {
  ```

  So `delegates` contains only the single matching row.
- Rank/rate are then assigned by array index over that filtered result (`modules/delegates.js:473-476`):

  ```js
  for (var i = 0; i < delegates.length; i++) {
    delegates[i].rate = i + 1; // always 1 when only one row was fetched
    delegates[i].rank = i + 1; // always 1 when only one row was fetched
    ...
  }
  ```

  With a one-element array, `i` is always `0`, so `rank`/`rate` are always `1`. `approval` and `productivity` are computed per-row and remain correct, which is why only `rank`/`rate` are wrong.

For comparison, the pre-regression `getDelegate` fetched the **full** delegate list (no filter), computed rank/rate over the whole sorted set, and only then selected the matching delegate with `_.find(...)`. That is why the old node returns the correct `rank: 82`.

### Proposed fix

Compute `rank`/`rate` over the full ordered delegate set, then apply the single-delegate selection. Options (smallest-safe-change preferred per `AGENTS.md`):

- Keep the SQL-level `filter` only for narrowing which row is **returned**, but compute the rank against the full ordered list; or
- Restore full-list ranking inside `getDelegate` and select the requested delegate after rank/rate are assigned (as before `60d7733c`), keeping the single-delegate convenience of the endpoint.

Whichever path is chosen, the other `delegates/get` lookup keys (`publicKey`, `address`, `username`) must all return the correct rank, and the existing `GET /api/delegates` list endpoint behavior must stay unchanged. `rate` remains a deprecated alias of `rank` (see the `TODO` at `modules/delegates.js:474`) and must keep matching `rank` until it is removed.

### Tests to add

There is currently **no** dedicated test for `GET /api/delegates/get` in `test/api/delegates.js` (it has blocks for `GET /api/delegates`, `/count`, `/voters`, `/search`, etc., but not `/get`). That coverage gap is what allowed the regression to ship. Add regression coverage:

- API tests (`test/api/delegates.js`): a new `describe('GET /api/delegates/get')` block asserting that a known delegate's `rank`/`rate` from `delegates/get` equals the rank/rate reported for the same delegate by the full `GET /api/delegates` list, and is **not** hard-coded to `1`. Cover all three lookup keys: `username`, `publicKey`, `address`.
- Unit tests (`test/unit/modules/delegates.js`): assert that `getDelegates(query, filter, cb)` returns the correct global `rank`/`rate` for a delegate that is not first in the ordering.

### Environment

- Affected: current node release line (`v0.9.0`+, including `endless.adamant.im`).
- Not affected: pre-`v0.9.0` nodes (e.g. `bid.adamant.im`).
- Endpoint: `GET /api/delegates/get` (handler `modules/delegates.js` `getDelegate` / `getDelegates`).

### References

- Regression commit: `60d7733c` `feat: find single delegate by the given address`
- Relevant code: `modules/delegates.js:442` (`getDelegates`), `modules/delegates.js:784` (`getDelegate`)
- API schema: https://github.com/Adamant-im/adamant-schema (verify `delegates/get` response contract once fixed)

massivedev0:
## Description

`GET /api/delegates/get` returned `rank: 1` and `rate: 1` for **every** delegate, regardless of the delegate's real position in the ranking. Regression introduced in `60d7733c` (`feat: find single delegate by the given address`, shipped in `v0.9.0`): the per-delegate `filter` was pushed into the accounts query, so `rank`/`rate` — assigned by array index over the result — were computed against a single pre-filtered row and always resolved to `1`. The same single-row defect also mis-reported `productivity` for outsider delegates (rank > `slots.delegates`).

Comparison on a known delegate:

- Old node (correct): https://bid.adamant.im/api/delegates/get?username=kabadas → `"rate":82,"rank":82`
- Current node (buggy): https://endless.adamant.im/api/delegates/get?username=kabadas → `"rate":1,"rank":1`

### Fix

- **`modules/delegates.js`** — `getDelegate` now requests the full ordered delegate list (empty filter) and selects the requested delegate with `.find(...)`, so `rank`/`rate` are computed over all delegates. This restores the pre-regression behavior while keeping the address-lookup capability added in `60d7733c`. Selection matches `address` (case-insensitive, mirroring `Account.getAll`) or `username`; the existing `publicKey`/`address` mismatch guard is untouched. `getDelegates` and the `GET /api/delegates` list endpoint are unchanged.

### Tests

- **`test/api/delegates.js`** — new `GET /api/delegates/get` suite. The endpoint previously had **no** test coverage, which is how the regression shipped. Verifies `username`/`publicKey`/`address` lookups return the real `rank`, that `rate === rank` (deprecated alias), and `rank > 1` (regression guard), plus no-criteria and unknown-delegate error cases.
- **`test/unit/modules/delegates.js`** — added a rank/rate regression test (selects a non-first delegate and asserts its real rank), and enabled the previously commented-out address-lookup test (now passing thanks to the fix).

## Related issue

Closes #252

## How to test

```
npm run test:single -- test/unit/modules/delegates.js
```

With a local testnet running in parallel:

```
npm run start:testnet
npm run test:single -- test/api/delegates.js --grep "delegates/get"
```

- Unit: `28 passing`, including the new rank/rate test and the enabled address-lookup test.
- API: `8 passing`. Live check — `delegates/get?username=<lowest-ranked>` returned `"rate":101,"rank":101` (its real position), not `1`.

Lint clean on touched files: `npx eslint modules/delegates.js test/api/delegates.js test/unit/modules/delegates.js`. Note: this repo uses a flat `eslint.config.js`, so eslint is run without the legacy `ESLINT_USE_FLAT_CONFIG=false` flag.

Not run: full `npm run test:unit` / `npm run test:api` suites — the change is local to one read-only API handler and both affected test files pass.

## Risk areas

- **Consensus / protocol:** none. This is a read-only public API handler. Block/transaction/round logic, delegate-list ordering for forging (`generateDelegateList`), serialization, IDs, signatures, rewards, and activation gating are all untouched.
- **Behavior change:** `delegates/get` now returns the correct `rank`/`rate` (and correct `productivity` for outsider delegates). It fetches the full delegate list instead of a single pre-filtered row — the same data source used before the regression.

## Checklist

- [x] No security checks weakened.
- [x] Schema/SQL unaffected.
- [x] Tests added and run (targeted unit + API).

### DIFF
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -805,12 +805,23 @@ Delegates.prototype.shared = {
         return setImmediate(cb, 'Delegate publicKey does not match address');
       }
 
-      modules.delegates.getDelegates(req.body, filter, function (err, data) {
+      // `rank` and `rate` are derived from the full, ordered delegate list, so
+      // ranking must be computed over all delegates. We request the ranked list
+      // (empty filter) and then select the requested delegate, instead of
+      // pushing `filter` into the accounts query: a single-row result is always
+      // ranked first and would report `rank`/`rate` (and outsider productivity)
+      // as 1 for every delegate.
+      modules.delegates.getDelegates(req.body, {}, function (err, data) {
         if (err) {
           return setImmediate(cb, err);
         }
 
-        var delegate = data.delegates[0];
+        var delegate = data.delegates.find(function (delegate) {
+          if (filter.address) {
+            return delegate.address.toUpperCase() === filter.address.toUpperCase();
+          }
+          return delegate.username === filter.username;
+        });
 
         if (delegate) {
           return setImmediate(cb, null, { delegate: delegate });
--- a/test/api/delegates.js
+++ b/test/api/delegates.js
@@ -424,6 +424,79 @@ describe('GET /api/delegates', function () {
   });
 });
 
+describe('GET /api/delegates/get', function () {
+  // A delegate that is NOT ranked first, so the test catches the regression
+  // where `delegates/get` reported rank/rate as 1 for every single-delegate
+  // lookup (because ranking was computed over a one-row, pre-filtered result).
+  var referenceDelegate;
+
+  before(function (done) {
+    node.get('/api/delegates?orderBy=rank:asc', function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.property('delegates').that.is.an('array');
+      // The list is ordered by rank ascending, so the last entry has the
+      // highest rank number, guaranteeing rank > 1.
+      referenceDelegate = res.body.delegates[res.body.delegates.length - 1];
+      node.expect(referenceDelegate).to.have.property('rank').that.is.above(1);
+      done();
+    });
+  });
+
+  it('using username should return the delegate with its real rank and rate', function (done) {
+    node.get('/api/delegates/get?username=' + encodeURIComponent(referenceDelegate.username), function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.property('delegate').that.is.an('object');
+      node.expect(res.body.delegate.username).to.equal(referenceDelegate.username);
+      node.expect(res.body.delegate.rank).to.equal(referenceDelegate.rank);
+      // `rate` is a deprecated alias of `rank` and must keep matching it.
+      node.expect(res.body.delegate.rate).to.equal(referenceDelegate.rank);
+      // Regression guard: rank/rate must reflect the real position, not 1.
+      node.expect(res.body.delegate.rank).to.be.above(1);
+      done();
+    });
+  });
+
+  it('using publicKey should return the delegate with its real rank and rate', function (done) {
+    node.get('/api/delegates/get?publicKey=' + referenceDelegate.publicKey, function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.property('delegate').that.is.an('object');
+      node.expect(res.body.delegate.publicKey).to.equal(referenceDelegate.publicKey);
+      node.expect(res.body.delegate.rank).to.equal(referenceDelegate.rank);
+      node.expect(res.body.delegate.rate).to.equal(referenceDelegate.rank);
+      node.expect(res.body.delegate.rank).to.be.above(1);
+      done();
+    });
+  });
+
+  it('using address should return the delegate with its real rank and rate', function (done) {
+    node.get('/api/delegates/get?address=' + encodeURIComponent(referenceDelegate.address), function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.property('delegate').that.is.an('object');
+      node.expect(res.body.delegate.address).to.equal(referenceDelegate.address);
+      node.expect(res.body.delegate.rank).to.equal(referenceDelegate.rank);
+      node.expect(res.body.delegate.rate).to.equal(referenceDelegate.rank);
+      node.expect(res.body.delegate.rank).to.be.above(1);
+      done();
+    });
+  });
+
+  it('using no criteria should fail', function (done) {
+    node.get('/api/delegates/get', function (err, res) {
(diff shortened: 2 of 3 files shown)


dev-adamant-im:
Review summary for PR #253:

What looks good:
- The PR matches issue #252: `GET /api/delegates/get` again computes `rank`/`rate` against the full ordered delegate set and only then selects the requested delegate.
- The fix preserves the address lookup added in the regression commit, including the existing `publicKey`/`address` mismatch guard.
- The change is limited to a read-only public API handler and regression tests; it does not touch block/transaction serialization, delegate forging order, rounds, rewards, signatures, activation gates, or SQL state.
- The new API and unit coverage exercises username, publicKey, and address lookups and guards against the previous `rank: 1`/`rate: 1` behavior.

What to watch:
- `delegates/get` now intentionally reads the full ranked delegate list, matching the old correct behavior and the public list endpoint's data source. That is acceptable for this fix, but it is still worth keeping an eye on endpoint cost if the registered delegate set grows substantially.

Findings:
- No significant correctness, consensus-safety, security, or regression findings found. I did not leave inline comments because I did not find an evidence-backed line-level issue.

Validation run:
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npm run test:single -- test/unit/modules/delegates.js` -> 28 passing.
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npx eslint modules/delegates.js test/api/delegates.js test/unit/modules/delegates.js` -> clean.
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npm run start:testnet`, then `PATH=/opt/homebrew/opt/node@22/bin:$PATH npm run test:single -- test/api/delegates.js --grep "delegates/get"` -> 8 passing. The testnet was stopped with Ctrl+C and completed graceful cleanup.
- `git diff --check origin/dev...HEAD` -> clean.

Note: running the unit test under the default `node` on this machine fails before test execution because `libpq` is compiled for `NODE_MODULE_VERSION 127` while default Node v26 requires `147`; rerunning with the repo-compatible Node 22 passes.

adamantmm (approved):
lgtm

Issue #218, opened 2026-05-09
[Feat] Harden node anti-abuse limits for REST, peer, and socket traffic


adamant-al:
## Summary

Increase ADAMANT Node anti-abuse protection for public REST API, peer HTTP API, node-to-node WebSocket traffic, and client WebSocket subscriptions by updating disabled/legacy default request limits to modern, measured, operator-configurable safeguards.

The goal is to protect public nodes from abusive traffic while keeping honest node sync, block/transaction propagation, forging, explorers, wallets, and normal client subscriptions reliable.

## Motivation

Initial note:

> Why don't we set default values for peers.options.limits and api.options.limits to protect nodes from abuse?
>
> It was like this for more than 7 years according to git history. The options are in legacy mode and should probably be updated to match last version of express-rate-limiter.

ADAMANT nodes are community-operated infrastructure. Default anti-abuse behavior should be safe enough for independent operators, but not so strict that it harms decentralization by forcing operators to add custom reverse-proxy rules before running a reliable public node.

### Current code findings

- `helpers/request-limiter.js` mounts `express-rate-limit` and `express-slow-down` on `/api/` and `/peer/`.
- The current limiter defaults are effectively disabled: `skip` returns `true`, `delayMs = 0`, `delayAfter = 0`, and `windowMs = 60000`.
- `config.default.json` keeps both `api.options.limits` and `peers.options.limits` at `max: 0`, `delayMs: 0`, `delayAfter: 0`, `windowMs: 60000`.
- `schema/config.js` requires only the legacy limit fields: `max`, `delayMs`, `delayAfter`, and `windowMs`.
- `app.js` applies the request limiter before the body parsers and later caps raw, URL-encoded, and JSON request bodies at `2mb`; URL-encoded payloads also use `parameterLimit: 5000`.
- The node uses `express-rate-limit = 7.5.0`, `express-slow-down = 2.1.0`, `express = 5.1.0`, `socket.io = 4.8.1`, and `body-parser = 2.2.0`.
- `npm view` currently reports `express-rate-limit = 8.5.1` and `express-slow-down = 3.1.0`.
- Current `express-rate-limit` docs use `limit` as the primary option, with `standardHeaders`, `legacyHeaders`, `identifier`, `requestPropertyName`, `validate`, `keyGenerator`, and store options. `max` is legacy-compatible but should not be the only schema field for new config.
[...]

PR #263, opened 2026-07-14
Fix: Optimize blocks API queries

### LINKED ISSUE #258: [Bug] blocks?generatorPublicKey with an unknown key scans the whole table and times out
## Summary

`GET /api/blocks?generatorPublicKey=…&orderBy=height:desc` stalls the node for tens of seconds when the given key has forged no or few blocks. On public nodes behind nginx the request ends with a `504 Gateway Time-out`. A single cheap request occupies a DB backend for a full table scan, so this is also an easy resource-exhaustion vector.

## Steps to reproduce

1. Take any syntactically valid 64-character hex key that does not belong to an active delegate, for example `aa11111111111111111111111111111111111111111111111111111111111111`
2. Request:

```sh
curl 'https://endless.adamant.im/api/blocks?generatorPublicKey=aa11111111111111111111111111111111111111111111111111111111111111&orderBy=height:desc&limit=1'
```

3. Compare with the same request for an active delegate's `generatorPublicKey`, which returns quickly.

## Expected behavior

The request returns `{"success": true, "blocks": [], "count": 0}` within milliseconds: the key matches nothing, and the lookup should be bounded by an index.

## Actual behavior

The request runs until the reverse proxy cuts it off (`504 Gateway Time-out` after 60+ seconds). Reproduced on `endless.adamant.im`, `lake.adamant.im`, and `clown.adamant.im`; verified again on ADAMANT Node v0.10.0 (2026-07-07).

## Environment

- ADAMANT Node v0.10.0, mainnet public nodes
- Blocks table height at the time of testing: ~53.5M rows

## Analysis

Migration `sql/migrations/20221009164928_createGeneratorPublicKeyIndex.sql` adds a **hash** index:

```sql
CREATE INDEX IF NOT EXISTS "blocks_b_generator_public_key"
ON "blocks" USING HASH ("text_generatorPublicKey");
```

A hash index provides no ordering, so for `WHERE "text_generatorPublicKey" = $1 ORDER BY "height" DESC LIMIT n` the PostgreSQL planner tends to walk the `height` index backwards and filter rows by generator, expecting an early match. That is fast for an active delegate (a match appears within the first few hundred entries) and degenerates into a scan over the whole ~53M-row table for a key with no or few blocks. This is the classic `LIMIT` + `ORDER BY` planner trap.

Suggested fix: add a composite btree index and let the planner satisfy both the filter and the ordering from it:

```sql
CREATE INDEX IF NOT EXISTS "blocks_generator_public_key_height"
ON "blocks" ("text_generatorPublicKey", "height" DESC);
```

The hash index can be dropped afterwards, since the composite btree also serves plain equality lookups.

Verification: `EXPLAIN ANALYZE` shows an index scan on the new index for both known and unknown keys; the unknown-key request above returns an empty result in milliseconds; delegate pages that list last forged blocks keep working.

Related: [#218](https://github.com/Adamant-im/adamant/issues/218) (anti-abuse hardening) — this endpoint is currently a one-request DoS regardless of rate limits.

## Pull requests

- Node implementation: https://github.com/Adamant-im/adamant/pull/263
- Documentation: https://github.com/Adamant-im/docs/pull/33
- OpenAPI schema: https://github.com/Adamant-im/adamant-schema/pull/46

massivedev0:
## Description

Prevent `GET /api/blocks` from scanning the blocks table when a syntactically valid but unknown `generatorPublicKey` is combined with the default `height:desc` ordering.

The previous hash index could filter by generator but could not satisfy the ordering. PostgreSQL could therefore choose a backward scan of the height index and filter every block when the generator had no matches. This PR replaces it with a composite B-tree index on `(text_generatorPublicKey, height DESC)` and drops the obsolete hash index after the new index is created.

The blocks query audit also found that `numberOfTransactions=0` was ignored because the filter used a truthiness check. The filter and request schema now handle zero correctly.

Tests cover unknown generators, every supported equality filter, pagination, every allowed `orderBy` field, the zero-transaction edge case, and the resulting SQL index definitions.

## Related issue

Closes #258

## External links (optional)

- Related anti-abuse hardening: #218
- Documentation issue: https://github.com/Adamant-im/docs/issues/32
- Documentation PR: https://github.com/Adamant-im/docs/pull/33
- Schema issue: https://github.com/Adamant-im/adamant-schema/issues/45
- Schema PR: https://github.com/Adamant-im/adamant-schema/pull/46

Not applicable.

## Breaking changes

No API or consensus-breaking changes. The response contract remains unchanged.

The first startup after upgrading must build the new B-tree index. On mainnet databases with tens of millions of blocks, operators should allow additional migration time and disk space.

## How to test

1. Apply migrations to a PostgreSQL database containing blocks.
2. Request an unknown but valid generator key:

   ```sh
   curl 'http://localhost:36666/api/blocks?generatorPublicKey=aa11111111111111111111111111111111111111111111111111111111111111&orderBy=height:desc&limit=1'
   ```

3. Verify the response contains an empty `blocks` array without a timeout.
4. Run:

   ```sh
   PG_NATIVE=false npm run test:single -- test/unit/modules/blocks.api.js test/unit/sql/blocks.js
   PG_NATIVE=false npm run test:unit:fast
   npx eslint modules/blocks/api.js schema/blocks.js test/api/blocks.js test/unit/modules/blocks.api.js test/unit/sql/blocks.js
   git diff --check
   ```

Local validation results:

- Targeted unit and SQL tests: 13 passing.
- Fast unit suite: 918 passing.
- ESLint for touched JavaScript files: passed.
- `git diff --check`: passed.
- On a local 427,709-row blocks table, the unknown-generator query improved from approximately 9.48 seconds to 8.45 milliseconds.

The full blocks API suite was not completed because the local testnet began a long catch-up sync and serialized API access behind `dbSequence`. The node was stopped through the graceful shutdown path.

## Notes for reviewers

- The migration creates the composite index before removing the old hash index, inside one transaction.
- Existing indexes cover the remaining equality filters and sortable block fields.
- `count` is intentionally not restored to the `/api/blocks` response; it is absent from the current documented API contract and was previously removed for performance reasons.
- Consensus, block serialization, validation, and replay behavior are unchanged.

## Checklist

- [x] Code is formatted
- [x] Documentation reviewed; companion updates submitted separately
- [x] Changes tested in all relevant local environments available for this change

### DIFF
--- a/modules/blocks/api.js
+++ b/modules/blocks/api.js
@@ -97,8 +97,7 @@ __private.list = function (filter, cb) {
     params.generatorPublicKey = filter.generatorPublicKey;
   }
 
-  // FIXME: Useless condition
-  if (filter.numberOfTransactions) {
+  if (filter.numberOfTransactions >= 0) {
     where.push('"b_numberOfTransactions" = ${numberOfTransactions}');
     params.numberOfTransactions = filter.numberOfTransactions;
   }
--- a/schema/blocks.js
+++ b/schema/blocks.js
@@ -40,6 +40,10 @@ module.exports = {
         type: 'string',
         format: 'publicKey'
       },
+      numberOfTransactions: {
+        type: 'integer',
+        minimum: 0
+      },
       totalAmount: {
         type: 'integer',
         minimum: 0,
--- a/sql/migrations/20260714120000_createBlocksGeneratorPublicKeyHeightIndex.sql
+++ b/sql/migrations/20260714120000_createBlocksGeneratorPublicKeyHeightIndex.sql
@@ -1,0 +1,12 @@
+/*
+ * Support generatorPublicKey filtering together with the default height order.
+ */
+
+BEGIN;
+
+CREATE INDEX IF NOT EXISTS "blocks_generator_public_key_height"
+ON "blocks" ("text_generatorPublicKey", "height" DESC);
+
+DROP INDEX IF EXISTS "blocks_b_generator_public_key";
+
+COMMIT;
--- a/test/api/blocks.js
+++ b/test/api/blocks.js
@@ -8,6 +8,9 @@ var block = {
   blockHeight: 0,
   id: 0,
   generatorPublicKey: '',
+  numberOfTransactions: 0,
+  previousBlock: '',
+  reward: 0,
   totalAmount: 0,
   totalFee: 0
 };
@@ -248,6 +251,9 @@ describe('GET /blocks', function () {
       node.expect(res.body.blocks[0].height).to.equal(block.blockHeight);
       block.id = res.body.blocks[0].id;
       block.generatorPublicKey = res.body.blocks[0].generatorPublicKey;
+      block.numberOfTransactions = res.body.blocks[0].numberOfTransactions;
+      block.previousBlock = res.body.blocks[0].previousBlock;
+      block.reward = res.body.blocks[0].reward;
       block.totalAmount = res.body.blocks[0].totalAmount;
       block.totalFee = res.body.blocks[0].totalFee;
       done();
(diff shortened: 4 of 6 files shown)


adamantmm (approved):
🚀 much faster now

PR #264, opened 2026-07-14
Feat: Add client WebSocket block and balance events

### LINKED ISSUE #217: [Feat] Allow clients to subscribe to balance and unconfirmedBalance changes
## Summary

Allow client applications connected through the node client WebSocket API to opt in to account balance updates for subscribed addresses, including `balance` and `unconfirmedBalance`.

## Motivation

ADAMANT clients can already subscribe to address-related transactions over Socket.IO, but they still need extra REST polling to know when an address balance or unconfirmed balance changed. This is especially useful for wallets and messaging clients that need to react quickly after a transaction enters the unconfirmed pool, gets confirmed, rolls back, or expires.

Adding an optional balance subscription would reduce polling, make client state fresher, and keep the feature additive for existing WebSocket consumers.

## Detailed description

Current client WebSocket behavior, based on a quick code review:

- `modules/clientWs.js` creates a Socket.IO server from `config.wsClient`.
- Connected clients can subscribe with socket events:
  - `address`: one address or an array of addresses.
  - `types`: one transaction type or an array of transaction types.
  - `assetChatTypes`: one chat asset type or an array of chat asset types.
- Subscriptions are stored in `modules/clientWs/transactionSubscription.js`.
- Matching transactions are emitted as `newTrans`.
- `logic/transaction.js` calls `clientWs.emit(new_trs)` from `Transaction.prototype.applyUnconfirmed()`.
- There is no current client WebSocket subscription for `balance` or `unconfirmedBalance`.
- REST account APIs already expose the fields:
  - `modules/accounts.js` returns `balance` and `unconfirmedBalance` in `getBalance`.
  - `modules/accounts.js` returns `account.balance` and `account.unconfirmedBalance` in account responses.

Requested behavior:

- A client should be able to subscribe to balance-related changes for one or more ADAMANT addresses.
- The subscription should support `balance`, `unconfirmedBalance`, or both.
- The feature must be opt-in so existing clients that only need `newTrans` do not trigger extra account lookups or receive extra events.
- The pushed payload should use public API naming:

```json
{
  "address": "U123456789",
  "balance": "100000000",
  "unconfirmedBalance": "99900000"
}
```

The exact socket event names should be decided during implementation. One possible shape:

```js
socket.emit('address', 'U123456789');
socket.emit('balances', ['balance', 'unconfirmedBalance']);

// Node emits only for subscribed addresses and requested fields:
socket.on('balances/change', (payload) => {});
```

Alternative shape:

```js
socket.emit('account', {
  addresses: ['U123456789'],
  fields: ['balance', 'unconfirmedBalance'],
  includeInitial: true
});

socket.on('account/change', (payload) => {});
```

## Screenshots or videos

Not applicable.

## Alternatives

- Keep client-side REST polling through `/api/accounts/getBalance` or account endpoints. This is simple, but it is wasteful and slower than the existing WebSocket path.
- Extend the existing `newTrans` transaction payload with balance fields. This is less explicit and may be surprising for clients that only subscribed to transaction events.
- Emit balance changes for every address subscription by default. This should be avoided because it changes the WebSocket contract and may add unnecessary DB/account lookups.

Recommended low-risk direction:

1. Extend the client WebSocket subscription model with an explicit set of account fields, limited to `balance` and `unconfirmedBalance`.
2. Keep existing `address`, `types`, `assetChatTypes`, and `newTrans` behavior unchanged.
3. Add validation for requested account fields and reuse existing address validation from `TransactionSubscription`.
4. Emit balance updates only for sockets that:
   - subscribed to the affected address;
   - explicitly requested `balance` and/or `unconfirmedBalance`.
5. Emit only after the account state mutation succeeds, so pushed values reflect current `mem_accounts` state.
6. Fetch current values through existing account accessors and serialize them with the same public names used by REST APIs.
7. Do not let WebSocket emission failure affect transaction/block processing. Any emit errors should be caught and logged only.

Implementation notes from the current code:

- `Transaction.prototype.applyUnconfirmed()` currently calls `clientWs.emit(new_trs)` before `u_balance` is merged. A new `unconfirmedBalance` notification should not reuse that exact timing if the payload is supposed to contain the updated value.
- Confirmed `balance` changes happen in `Transaction.prototype.apply()`/`undo()` and in type-specific apply/undo handlers for recipients, such as `logic/transfer.js` and `logic/chat.js`.
- For complete coverage, implementation should identify all account balance mutation paths, including block apply/undo, unconfirmed apply/undo, transaction expiration/removal, and round/reward updates. If full coverage is too broad for the first PR, document the first supported cases explicitly.
- This should be treated as a node client API feature, not a consensus or wire-format change. It must not alter transaction bytes, IDs, signatures, validation, block replay behavior, or balance mutation ordering.

Suggested tests:

- Unit tests for new subscription validation in `test/unit/modules/clientWs.js`.
- Tests that legacy `address`/`types`/`assetChatTypes` matching remains unchanged.
- Tests that invalid balance field names are ignored.
- Tests that balance events are emitted only for subscribed addresses and requested fields.
- Integration or module-level tests for unconfirmed transaction flow showing that `unconfirmedBalance` is emitted after the account value changes.

Documentation/schema follow-up:

- Update node docs for `wsClient` subscriptions.
- Update `adamant-schema` if the WebSocket payloads are represented there.
- Mention that `unconfirmedBalance` is mempool/live-node state and can change on transaction expiration, rollback, or revalidation.

## Acceptance criteria

- [ ] Clients can opt in to `balance` and `unconfirmedBalance` updates for specific addresses over the client WebSocket API.
- [ ] Existing transaction subscriptions and `newTrans` payloads remain backward-compatible.
- [ ] Invalid addresses and unsupported balance fields are ignored or rejected consistently with current subscription validation style.
- [ ] Balance payloads use string numeric values and public names matching REST APIs.
- [ ] WebSocket emit failures cannot break account mutation, transaction processing, block processing, or replay.
- [ ] Tests cover subscription validation, opt-in behavior, and at least one unconfirmed balance update flow.
- [ ] Docs/schema follow-up is completed or tracked in linked issues.

## Related context

- Related but different: #201 discusses returning `balance` and `unconfirmedBalance` in chat/chatroom API responses for anti-spam sender data.
- This issue is specifically about real-time client WebSocket subscriptions for account balance changes.

## Pull request

- https://github.com/Adamant-im/adamant/pull/264

### LINKED ISSUE #256: [Feat] Emit newBlock events over the client WebSocket API
## Summary

Add an opt-in `newBlock` subscription to the client WebSocket API so block-oriented consumers can receive new blocks as push events instead of polling REST endpoints.

## Motivation

The client WebSocket API (`modules/clientWs.js`) currently emits only per-transaction `newTrans` events. Consumers that need block-level data have to poll REST aggressively:

- ADAMANT Explorer polls `/api/blocks` every second for the Delegate Monitor and refreshes the full delegate state every 5 seconds, see [Adamant-im/adamant-explorer#11](https://github.com/Adamant-im/adamant-explorer/issues/11)
- Forging pools and monitoring tools follow the same pattern to detect new blocks and missed slots

A `newBlock` push event would remove most of this polling, reduce REST load on public nodes, and make block-driven UIs update as soon as a block is applied. This goes in the same direction as #217, which proposes balance subscriptions over the same channel.

## Detailed description

Current client WebSocket behavior, based on a code review:

- `modules/clientWs.js` creates a Socket.IO server from `config.wsClient`
- Connected clients subscribe with socket events `address`, `types`, and `assetChatTypes`
- Matching transactions are emitted as `newTrans`; emission is deduplicated by transaction id
- `logic/transaction.js` calls `clientWs.emit(new_trs)` from `Transaction.prototype.applyUnconfirmed()`
- The peer transport (`modules/transport.js`) has an internal `blocks/change` signal, but it is not exposed to client WebSocket consumers

Requested behavior:

- A client can opt in to block events, for example with `socket.emit('blocks', true)`
- On every applied block, the node emits `newBlock` to subscribed clients with the block header in public API naming:

```json
{
  "id": "1739739671268673627",
  "height": 53532078,
  "timestamp": 278966175,
  "generatorPublicKey": "24636735d64711f91f7680bb348662bc74f7b3446e66702369527e3882dd30d6",
  "numberOfTransactions": 2,
  "totalAmount": 10000000,
  "totalFee": 100000,
  "reward": 10000000
}
```

- The subscription must be opt-in so existing `newTrans` consumers see no new traffic
- Exact event and payload names can be decided during implementation

## Alternatives

- Keep REST polling: works today but wastes node resources and adds up to a block interval of latency
- Expose the peer transport `blocks/change` signal directly: mixes the peer protocol with the client API and is not opt-in

- Extend the subscription state in `modules/clientWs.js` with a per-socket block flag, next to `address`/`types`/`assetChatTypes`
- Emit from the block application path in the blocks chain module, mirroring how `logic/transaction.js` calls `clientWs.emit()`
- Deduplicate by block id the same way `newTrans` deduplicates by transaction id
- Cover with unit tests next to `test/unit/modules/clientWs.js`

Verification: connect a Socket.IO client to `portWS`, subscribe to blocks, and observe `newBlock` events matching the `/api/blocks/getHeight` progression (one event per ~5 seconds).

## Pull request

- https://github.com/Adamant-im/adamant/pull/264

massivedev0:
## Description

Add opt-in client Socket.IO subscriptions for new blocks and account balance changes.

- `socket.emit('blocks', true)` enables compact `newBlock` events after successful block application.
- `address` plus `balances` subscriptions enable `balances/change` events for `balance`, `unconfirmedBalance`, or both.
- Balance payloads use public REST field names and decimal strings.
- Block, transaction, and balance subscriptions use dedicated indexes so unrelated clients are not scanned and account reads are skipped when nobody subscribed to the changed address and field.
- Balance changes during block apply or rollback are batched into one final account read per address.
- Socket and event-publication failures are isolated from account, transaction, round, and block processing.

The existing `address`, `types`, `assetChatTypes`, and `newTrans` contracts remain backward-compatible.

## Related issue

Closes #217
Closes #256

## External links (optional)

- Documentation follow-up: Adamant-im/docs#34 and Adamant-im/docs#35
- Schema follow-up: Adamant-im/adamant-schema#47 and Adamant-im/adamant-schema#48

Not applicable. This is a Socket.IO API and node-internal publication change.

## Breaking changes

None. Both new event families are opt-in. Existing transaction subscribers receive no additional traffic unless they explicitly subscribe to blocks or balance fields.

## How to test

Environment checks:

```sh
pg_isready -h localhost -p 5432
redis-cli -h 127.0.0.1 -p 6379 ping
```

Targeted WebSocket, account, block, and round tests:

```sh
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run test:single -- \
  test/unit/modules/clientWs.js \
  test/unit/modules/clientWs.server.js \
  test/unit/modules/clientWs.integration.js \
  test/unit/logic/account.js \
  test/unit/logic/transaction.js \
  test/unit/modules/blocks.js \
  test/unit/modules/rounds.js
```

Expected result: `226 passing`.

Broader fast unit coverage and lint:

```sh
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run test:unit:fast
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run eslint
git diff --check origin/dev...HEAD
```

Expected results: `938 passing`, ESLint `Done.`, and no diff errors.

The full non-fast unit suite, local testnet, and API suite were not run because the change does not alter consensus validation, serialization, SQL, peer transport, or REST endpoints, and the directly affected paths are covered by unit and real Socket.IO integration tests.

## Notes for reviewers

- `newBlock` is emitted only after the block pipeline succeeds and only when the block is being saved; database replay does not create client events.
- Account events are scheduled only after a successful account merge. Round completion explicitly publishes affected delegate balance fields.
- Block apply and rollback use nested balance batches to avoid exposing transient unconfirmed-pool rewind state.
- The balance subscription index prevents account database reads when no socket subscribed to the affected address and field.
- Duplicate transaction and block IDs are suppressed for at least 60 seconds; periodic cleanup can extend the effective window to approximately two minutes.
- Event delivery is best-effort and non-durable; clients should restore subscriptions after reconnecting and reconcile state through REST.

## Risk areas

- **Consensus and protocol:** No transaction or block bytes, IDs, signatures, validation rules, activation gates, reward calculations, delegate ordering, or slot timing are changed.
- **Replay and storage:** No SQL schema, migration, or persisted model changes. Historical database replay does not emit `newBlock`.
- **Security and privacy:** Existing client WebSocket exposure is preserved. Inputs remain allowlisted, no authentication or peer-handshake behavior changes, and no sensitive values are logged.
- **Performance:** Subscription indexes avoid unrelated scans and account reads; block apply/rollback coalesces balance notifications into final-state reads.

## Checklist

- [x] Code is formatted
- [x] Documentation updated (if needed)
- [x] Changes tested in all relevant environments

### DIFF
--- a/logic/account.js
+++ b/logic/account.js
@@ -986,7 +986,11 @@ Account.prototype.merge = function (address, diff, cb) {
       return Object.prototype.hasOwnProperty.call(update, field);
     });
 
-    if (changedBalanceFields.length && account.scope.clientWs) {
+    if (
+      changedBalanceFields.length &&
+      account.scope.clientWs &&
+      typeof account.scope.clientWs.emitBalanceChange === 'function'
+    ) {
       try {
         account.scope.clientWs.emitBalanceChange(
             address,
--- a/logic/account.js
+++ b/logic/account.js
@@ -22,13 +22,20 @@ var self, library, __private = {};
  * @param {Database} db
  * @param {ZSchema} schema
  * @param {object} logger
+ * @param {ClientWs} [clientWs] - Optional client WebSocket event publisher.
  * @param {function} cb - Callback function.
  * @return {setImmediateCallback} With `this` as data.
  */
-function Account (db, schema, logger, cb) {
+function Account (db, schema, logger, clientWs, cb) {
+  if (typeof clientWs === 'function') {
+    cb = clientWs;
+    clientWs = null;
+  }
+
   this.scope = {
     db: db,
-    schema: schema
+    schema: schema,
+    clientWs: clientWs
   };
 
   self = this;
@@ -750,6 +757,7 @@ Account.prototype.set = function (address, rawFields, cb) {
  */
 Account.prototype.merge = function (address, diff, cb) {
   var update = {}, remove = {}, insert = {}, insert_object = {}, remove_object = {}, round = [];
+  var account = this;
 
   // Verify public key
   this.verifyPublicKey(diff.publicKey);
@@ -974,6 +982,26 @@ Account.prototype.merge = function (address, diff, cb) {
   }
 
   this.scope.db.none(queries).then(function () {
+    const changedBalanceFields = ['balance', 'u_balance'].filter(function (field) {
+      return Object.prototype.hasOwnProperty.call(update, field);
+    });
+
+    if (changedBalanceFields.length && account.scope.clientWs) {
+      try {
+        account.scope.clientWs.emitBalanceChange(
+            address,
+            changedBalanceFields,
+            account.get.bind(account, { address: address }, ['address', 'balance', 'u_balance'])
+        );
+      } catch (err) {
+        library.logger.debug(
+            'ws-client-server',
+            `Unable to publish balance change for ${address}: ${err?.message || err}`,
+            err?.stack
+        );
+      }
+    }
+
     return done();
   }).catch(function (err) {
     library.logger.error('account', `An error occurred while trying to merge account data: ${err?.message || err}`, err.stack);
--- a/modules/blocks/chain.js
+++ b/modules/blocks/chain.js
@@ -340,7 +340,8 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
   var balanceBatchStarted = false;
   if (
     library.clientWs &&
-    typeof library.clientWs.beginBalanceBatch === 'function'
+    typeof library.clientWs.beginBalanceBatch === 'function' &&
+    typeof library.clientWs.endBalanceBatch === 'function'
   ) {
     try {
       library.clientWs.beginBalanceBatch();
@@ -500,7 +501,7 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
     function finalize (err) {
       if (balanceBatchStarted) {
         try {
-          library.clientWs.endBalanceBatch();
+          library.clientWs.endBalanceBatch(!err && Boolean(saveBlock));
         } catch (batchErr) {
           library.logger.debug(
               'ws-client-server',
@@ -510,7 +511,12 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
         }
       }
 
-      if (!err && saveBlock && library.clientWs) {
+      if (
+        !err &&
+        saveBlock &&
+        library.clientWs &&
+        typeof library.clientWs.emitBlock === 'function'
+      ) {
         try {
           library.clientWs.emitBlock(block);
         } catch (emitErr) {
@@ -657,7 +663,8 @@ Chain.prototype.deleteLastBlock = function (cb) {
   var balanceBatchStarted = false;
   if (
     library.clientWs &&
-    typeof library.clientWs.beginBalanceBatch === 'function'
+    typeof library.clientWs.beginBalanceBatch === 'function' &&
+    typeof library.clientWs.endBalanceBatch === 'function'
   ) {
     try {
       library.clientWs.beginBalanceBatch();
@@ -675,7 +682,7 @@ Chain.prototype.deleteLastBlock = function (cb) {
   __private.popLastBlock(lastBlock, function (err, newLastBlock) {
     if (balanceBatchStarted) {
       try {
-        library.clientWs.endBalanceBatch();
+        library.clientWs.endBalanceBatch(!err);
       } catch (batchErr) {
         library.logger.debug(
             'ws-client-server',
--- a/modules/blocks/chain.js
+++ b/modules/blocks/chain.js
@@ -22,14 +22,16 @@ var modules, library, self, __private = {};
  * @param {object} genesisblock
  * @param {bus} bus
  * @param {Sequence} balancesSequence
+ * @param {ClientWs} clientWs - Client WebSocket event publisher
  */
-function Chain (logger, block, transaction, db, genesisblock, bus, balancesSequence) {
+function Chain (logger, block, transaction, db, genesisblock, bus, balancesSequence, clientWs) {
   library = {
     logger: logger,
     db: db,
     genesisblock: genesisblock,
     bus: bus,
     balancesSequence: balancesSequence,
+    clientWs: clientWs,
     logic: {
       block: block,
       transaction: transaction
@@ -335,6 +337,23 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
   // Prevent shutdown during database writes.
   modules.blocks.isActive.set(true);
 
+  var balanceBatchStarted = false;
+  if (
+    library.clientWs &&
+    typeof library.clientWs.beginBalanceBatch === 'function'
+  ) {
+    try {
+      library.clientWs.beginBalanceBatch();
+      balanceBatchStarted = true;
+    } catch (batchErr) {
+      library.logger.debug(
+          'ws-client-server',
+          `Unable to begin balance event batch: ${batchErr?.message || batchErr}`,
+          batchErr?.stack
+      );
+    }
+  }
+
   // Transactions to rewind in case of error.
   var appliedTransactions = {};
 
@@ -479,6 +498,30 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
     }
   }, function (err) {
     function finalize (err) {
+      if (balanceBatchStarted) {
+        try {
+          library.clientWs.endBalanceBatch();
+        } catch (batchErr) {
+          library.logger.debug(
+              'ws-client-server',
+              `Unable to finish balance event batch: ${batchErr?.message || batchErr}`,
+              batchErr?.stack
+          );
+        }
+      }
+
+      if (!err && saveBlock && library.clientWs) {
+        try {
+          library.clientWs.emitBlock(block);
+        } catch (emitErr) {
+          library.logger.debug(
+              'ws-client-server',
+              `Unable to publish block ${block.id}: ${emitErr?.message || emitErr}`,
+              emitErr?.stack
+          );
+        }
+      }
+
       // Allow shutdown, database writes are finished.
       modules.blocks.isActive.set(false);
 
@@ -611,8 +654,37 @@ Chain.prototype.deleteLastBlock = function (cb) {
     return setImmediate(cb, 'Cannot delete genesis block');
   }
 
+  var balanceBatchStarted = false;
+  if (
+    library.clientWs &&
+    typeof library.clientWs.beginBalanceBatch === 'function'
+  ) {
+    try {
+      library.clientWs.beginBalanceBatch();
+      balanceBatchStarted = true;
+    } catch (batchErr) {
+      library.logger.debug(
+          'ws-client-server',
+          `Unable to begin rollback balance event batch: ${batchErr?.message || batchErr}`,
+          batchErr?.stack
+      );
+    }
+  }
+
   // Delete last block, replace last block with previous block, undo things
   __private.popLastBlock(lastBlock, function (err, newLastBlock) {
+    if (balanceBatchStarted) {
+      try {
+        library.clientWs.endBalanceBatch();
+      } catch (batchErr) {
+        library.logger.debug(
+            'ws-client-server',
+            `Unable to finish rollback balance event batch: ${batchErr?.message || batchErr}`,
+            batchErr?.stack
+        );
+      }
+    }
+
     if (err) {
       library.logger.error('blocks', 'An error occurred while deleting last block', lastBlock);
     } else {
--- a/modules/clientWs.js
+++ b/modules/clientWs.js
@@ -15,6 +15,7 @@ class ClientWs {
     this.blockSubscriptions = new Set();
     this.balanceSubscriptions = new Map();
     this.balanceBatchDepth = 0;
+    this.discardBalanceBatch = false;
     this.pendingBalanceChanges = new Map();
     this.logger = logger;
 
@@ -110,7 +111,7 @@ class ClientWs {
       let pending = this.pendingBalanceChanges.get(normalizedAddress);
 
       if (!pending) {
-        pending = { fields: new Set(), getAccount: getAccount };
+        pending = { fields: new Set() };
         this.pendingBalanceChanges.set(normalizedAddress, pending);
       }
 
@@ -159,17 +160,26 @@ class ClientWs {
    * @return {void}
    */
   beginBalanceBatch () {
+    if (this.balanceBatchDepth === 0) {
+      this.discardBalanceBatch = false;
+    }
+
     this.balanceBatchDepth += 1;
   }
 
   /**
    * Flushes one final balance read per changed address after the outer batch.
+   * Suppressing any nested batch discards the complete outer batch.
    * @param {boolean} [emitChanges=true] - Whether to publish or discard the batch
    * @return {void}
    */
   endBalanceBatch (emitChanges = true) {
     if (this.balanceBatchDepth === 0) {
       return;
+    }
+
+    if (!emitChanges) {
+      this.discardBalanceBatch = true;
     }
 
     this.balanceBatchDepth -= 1;
@@ -179,8 +189,10 @@ class ClientWs {
 
     const pendingChanges = this.pendingBalanceChanges;
     this.pendingBalanceChanges = new Map();
+    const discardBalanceBatch = this.discardBalanceBatch;
+    this.discardBalanceBatch = false;
 
-    if (!emitChanges) {
+    if (discardBalanceBatch) {
       return;
     }
 
--- a/modules/clientWs.js
+++ b/modules/clientWs.js
@@ -11,6 +11,11 @@ class ClientWs {
   constructor (config, logger, cb) {
     this.enabled = Boolean(config && config.enabled);
     this.describes = {};
+    this.transactionSubscriptions = new Set();
+    this.blockSubscriptions = new Set();
+    this.balanceSubscriptions = new Map();
+    this.balanceBatchDepth = 0;
+    this.pendingBalanceChanges = new Map();
     this.logger = logger;
 
     if (!this.enabled) {
@@ -22,6 +27,7 @@ class ClientWs {
       allowEIO3: true,
       cors: config.cors
     });
+    this.io = io;
 
     io.sockets.on('connection', this.handleConnection.bind(this));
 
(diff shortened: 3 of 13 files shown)


modules/blocks/chain.js
```diff
@@ -479,6 +498,30 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
     }
   }, function (err) {
     function finalize (err) {
+      if (balanceBatchStarted) {
+        try {
+          library.clientWs.endBalanceBatch();
```
adamantmm: **Replay/rebuild may still emit balance events**

`endBalanceBatch()` always runs in `finalize()` and flushes `pendingBalanceChanges`, regardless of `saveBlock`. During mem_* rebuild (`loadBlocksOffset` → `applyBlock(..., saveBlock: false)` in `modules/blocks/process.js:212`), account merges still happen and batched balance notifications are delivered to subscribed clients.

`newBlock` is correctly gated at line 513 (`!err && saveBlock`), but balance events are not.

**Recommendation:** mirror the `newBlock` gate — e.g. when `saveBlock` is false, skip `beginBalanceBatch()` entirely, or discard pending changes in `finalize()` without flushing. That keeps historical replay from generating live-looking `balances/change` traffic and avoids extra account reads during rebuild.

modules/blocks/chain.js
```diff
@@ -479,6 +498,30 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
...
+        } catch (batchErr) {
+          library.logger.debug(
+              'ws-client-server',
+              `Unable to finish balance event batch: ${batchErr?.message || batchErr}`,
+              batchErr?.stack
+          );
+        }
+      }
+
+      if (!err && saveBlock && library.clientWs) {
```
adamantmm: Good: gating `emitBlock` on `saveBlock` prevents replay from emitting `newBlock`. Please apply the same rule to balance publication (see comment on `endBalanceBatch()` above) so both event families behave consistently during rebuild.

logic/account.js
```diff
@@ -974,6 +982,26 @@ Account.prototype.merge = function (address, diff, cb) {
   }
 
   this.scope.db.none(queries).then(function () {
+    const changedBalanceFields = ['balance', 'u_balance'].filter(function (field) {
+      return Object.prototype.hasOwnProperty.call(update, field);
+    });
+
+    if (changedBalanceFields.length && account.scope.clientWs) {
```
adamantmm: Hooking balance publication into `merge()` after a successful DB write is the right integration point (covers confirmed/unconfirmed/undo paths and avoids the early `newTrans` timing in `applyUnconfirmed`).

One caveat: this path has no visibility into `saveBlock`, so replay can still publish if the batch is flushed in `chain.finalize()`. Fixing the batch gate in `chain.js` should be sufficient; no change needed here unless you want an explicit rebuild guard at the account layer.

adamantmm (commented):
## Review summary

Solid additive feature: opt-in `newBlock` and `balances/change`, good subscription indexes, balance batching during block apply/rollback, and socket failures are isolated from consensus paths. I ran the targeted suite from the PR description locally: **94 passing**, plus `npm run test:unit:fast`: **934 passing**.

No consensus, serialization, or validation changes observed. A few follow-ups below — the replay/saveBlock symmetry point is the only one I'd treat as merge-blocking or near-blocking.

**Verdict:** approve with minor follow-ups.

modules/clientWs.js
```diff
@@ -93,7 +229,63 @@ class ClientWs {
...
+      describe.addresses.size > 0 ||
+      describe.types.size > 0 ||
+      describe.assetChatTypes.size > 0
+    ) {
+      this.transactionSubscriptions.add(describe);
+    } else {
+      this.transactionSubscriptions.delete(describe);
+    }
+
+    if (describe.balanceFields.size > 0) {
```
adamantmm, line 297: **Balance index only grows on re-register**

When `balanceFields.size > 0`, subscriptions are added to `balanceSubscriptions`, but stale entries are never removed when `balanceFields` becomes empty (there is no unsubscribe API today, so this is not a live bug).

For symmetry with `blockSubscriptions` / `transactionSubscriptions` (lines 262–276), consider an `else` branch that removes `describe` from all indexed addresses when `balanceFields.size === 0`. That will matter if a future unsubscribe mechanism is added.

modules/clientWs.js
```diff
@@ -40,17 +46,139 @@ class ClientWs {
...
+      }
+
+      for (const field of publicFields) {
+        pending.fields.add(field);
+      }
+      pending.getAccount = getAccount;
+      return;
+    }
+
+    try {
```
adamantmm, line 125: Minor note: `getAccount` is async, so rapid successive merges for the same address could deliver `balances/change` out of order. The PR already documents best-effort delivery; worth a one-line comment here for future maintainers, or accepting it as documented behavior.

modules/rounds.js
```diff
@@ -373,6 +383,42 @@ Rounds.prototype.onFinishRound = function (round) {
   library.network.wsServer.emit('rounds/change', { number: round });
 };
 
+/**
+ * Publishes reward-related balance changes after a complete round transaction.
+ * Subscriber indexes prevent account reads for addresses without listeners.
+ * @param {object} scope - Completed forward or backward round scope
+ * @return {void}
+ */
+__private.emitRoundBalanceChanges = function (scope) {
```
adamantmm, line 392: **Test gap: round reward balance emission**

`emitRoundBalanceChanges` is new logic with delegate address derivation and `finishRound` gating, but there is no unit coverage in `test/unit/modules/rounds.js`.

**Recommendation:** add a focused test stubbing `library.clientWs.emitBalanceChange` and verifying it is called only when `finishRound` is true and `roundDelegates` is populated (and not called on intermediate blocks).

test/unit/logic/account.js
```diff
@@ -109,6 +109,52 @@ describe('account', () => {
       });
     });
 
+    it('should publish changed balance fields after a successful merge', (done) => {
```
adamantmm, line 112: Issue #217 asked for coverage of at least one unconfirmed balance update flow. These `merge()` tests are good, but they do not exercise the full `Transaction.prototype.applyUnconfirmed()` path where `newTrans` is still emitted before `u_balance` merge.

**Recommendation:** add a module-level test (e.g. in `clientWs.server.js` or a small transaction test) that applies an unconfirmed transaction and asserts `balances/change` arrives **after** `u_balance` is updated, with the correct string payload.

modules/clientWs.js
```diff
@@ -40,17 +46,139 @@ class ClientWs {
...
+          return;
+        }
+
+        for (const subscription of subscriptions) {
+          const payload = { address: normalizedAddress };
+
+          for (const field of publicFields) {
+            if (subscription.balanceFields.has(field)) {
+              const accountField = field === 'unconfirmedBalance' ? 'u_balance' : field;
+              payload[field] = String(account[accountField]);
```
al-onyxprotocol, line 143: **Balance is stringified from a JS `Number`, not a bigint-safe value.** `account.get` defines `balance`/`u_balance` with `conv: Number` (`logic/account.js:181`, `:193`), so `account[accountField]` is already a JS `Number` here. For balances above `2^53` base units (~90M ADM — genesis/exchange holders), the value is imprecise *before* `String(...)`, so `balances/change` can emit a wrong decimal string.

This matches the current REST representation, so it's not a regression, but the PR advertises "decimal strings," which implies exactness. **Recommendation:** confirm the emitted value is byte-for-byte identical to `/api/accounts` for a >2^53 balance, or read the field as text (`("balance")::text`) if exactness is intended.

modules/blocks/chain.js
```diff
@@ -335,6 +337,23 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
   // Prevent shutdown during database writes.
   modules.blocks.isActive.set(true);
 
+  var balanceBatchStarted = false;
+  if (
+    library.clientWs &&
+    typeof library.clientWs.beginBalanceBatch === 'function'
```
al-onyxprotocol: **Inconsistent capability guards.** The batch calls are guarded with `typeof library.clientWs.beginBalanceBatch === 'function'`, but `emitBlock` (line 515) and `emitBalanceChange` (in `account.js`/`rounds.js`) only check a truthy `library.clientWs`. In production `clientWs` is always a full `ClientWs`, so this is harmless, but the mixed convention reads as accidental. **Recommendation:** pick one convention — drop the `typeof` check here, or add it to the emit sites — so the defensiveness is uniform and intentional.

modules/clientWs.js
```diff
@@ -40,17 +46,139 @@ class ClientWs {
...
+      this.logError(`Unable to schedule balance lookup for ${normalizedAddress}`, e);
+    }
+  }
+
+  /**
+   * Defers balance reads while a block apply or rollback mutates transient state.
+   * Nested batches are supported because chain operations are explicitly sequenced.
+   * @return {void}
+   */
+  beginBalanceBatch () {
```
al-onyxprotocol, line 162: **Batch depth is process-global shared state.** `balanceBatchDepth`/`pendingBalanceChanges` live on the shared `ClientWs` instance, so correctness depends entirely on `applyBlock` and `deleteLastBlock` never overlapping. That's guaranteed today by `balancesSequence`, and the JSDoc notes it — good. But if that serialization assumption ever changes, depth accounting corrupts silently (mismatched begin/end across two concurrent operations). Non-blocking; just flagging the latent coupling. **Recommendation:** keep as-is with the comment, or make the invariant explicit (assert entry only from the sequence).

modules/clientWs.js
```diff
@@ -40,17 +46,139 @@ class ClientWs {
...
+    }
+  }
+
+  /**
+   * Emits a compact public block header to sockets that opted in to blocks.
+   * Historical rebuilds are filtered by the caller before reaching this method.
+   * @param {block} block - Successfully applied block
+   * @return {void}
+   */
+  emitBlock (block) {
```
al-onyxprotocol, line 72: **Optional readability nit:** the `newBlock` name is reused for both this client-facing socket event and the internal `library.bus.message('newBlock', ...)` in `chain.js`. Different channels, no functional collision, but easy to misread when tracing. A one-line comment here noting this is the public socket event (distinct from the internal bus message) would help future readers.

al-onyxprotocol (commented):
Reviewed the client WebSocket block/balance events. No blocking issues — the design is careful (subscriber indexes avoid unrelated scans and skip account reads with no listeners, apply/rollback coalesces balance changes into one final read per address, and every emit path is failure-isolated from consensus/account/block processing). All 45 targeted ClientWs tests pass locally.

A few minor, non-blocking remarks are left inline. Overall: **Approve.**

logic/account.js
massivedev0: Agreed; no account-layer replay guard was added. The fix is at the chain boundary in 15afe8d0, where replay and failed apply discard the batch before any balance lookup.

modules/clientWs.js
massivedev0: Clarified in 15afe8d0 that the public Socket.IO newBlock event is distinct from the internal bus event.

modules/rounds.js
massivedev0: Fixed in 15afe8d0. Added focused Rounds.tick coverage proving no balance publication on an intermediate block and one delegate balance publication after a completed round.

massivedev0:
Addressed the review feedback in `15afe8d0`.

- discard pending balance notifications during replay, failed block application, and failed rollback
- add completed-round versus intermediate-block balance publication coverage
- add full `applyUnconfirmed()` coverage proving the emitted getter observes updated `u_balance`
- make client WebSocket capability guards consistent
- document public/internal `newBlock` distinction and best-effort asynchronous balance ordering

Validation:

- targeted WebSocket/account/transaction/block/round suite: 226 passing
- `npm run test:unit:fast`: 938 passing
- `npm run eslint`: passed
- `git diff --check`: passed

The full non-fast unit suite and testnet/API suites were not rerun because this follow-up is covered by focused regression tests plus the fast shared-logic suite and does not change consensus validation, serialization, SQL, peer transport, or REST endpoints.

adamantmm (approved):

modules/clientWs.js
```diff
@@ -40,17 +46,146 @@ class ClientWs {
...
+   * @param {boolean} [emitChanges=true] - Whether to publish or discard the batch
+   * @return {void}
+   */
+  endBalanceBatch (emitChanges = true) {
+    if (this.balanceBatchDepth === 0) {
+      return;
+    }
+
+    this.balanceBatchDepth -= 1;
+    if (this.balanceBatchDepth > 0) {
```
al-onyxprotocol, line 186: **Latent: `emitChanges` is ignored for nested batches.** When depth is still > 0 after the decrement, the method returns before ever looking at `emitChanges`, so an inner `endBalanceBatch(false)` cannot discard anything — the outer flush will still emit those changes. Today `applyBlock` and `deleteLastBlock` never truly nest, and flushed payloads are re-read final state, so no wrong data can be emitted. But if nesting ever becomes real, the discard semantics silently break.

**Recommendation:** latch a `this.discardBalanceBatch ||= !emitChanges` flag on every call and honor it at the outermost flush, or assert/log when `emitChanges === false` is passed at depth > 1.

modules/clientWs.js
```diff
@@ -40,17 +46,146 @@ class ClientWs {
...
+
+    if (!this.hasBalanceSubscribers(normalizedAddress, publicFields)) {
+      return;
+    }
+
+    if (this.balanceBatchDepth > 0) {
+      let pending = this.pendingBalanceChanges.get(normalizedAddress);
+
+      if (!pending) {
+        pending = { fields: new Set(), getAccount: getAccount };
```
al-onyxprotocol: **Nit: redundant assignment.** `getAccount` is set here in the initializer and then unconditionally overwritten at line 120. The line-120 assignment (keep the newest accessor) is the meaningful one; the initializer copy is dead.

**Recommendation:** initialize with `pending = { fields: new Set(), getAccount: null };` or simply drop the property here and rely on line 120.

modules/clientWs.js
```diff
@@ -40,17 +46,146 @@ class ClientWs {
...
+   * This public Socket.IO event is distinct from the internal `newBlock` bus event.
+   * @param {block} block - Successfully applied block
+   * @return {void}
+   */
+  emitBlock (block) {
+    if (!this.enabled || !block || this.blockSubscriptions.size === 0) {
+      return;
+    }
+
+    if (lastBlocksIds.has(block.id)) {
```
al-onyxprotocol, line 77: **Edge case: dedup can suppress a legitimate re-apply.** If a block is deleted during fork resolution and the same block ID is re-applied within the ~60–120 s suppression window, the second `newBlock` event is silently dropped by `lastBlocksIds`. The consequence is only a missed notification and clients reconcile via REST, so this is fine as-is.

**Recommendation:** just document the behavior in the follow-up docs PR (Adamant-im/docs#34) alongside the existing best-effort caveats.

modules/blocks/chain.js
```diff
@@ -479,6 +499,35 @@ Chain.prototype.applyBlock = function (block, broadcast, cb, saveBlock) {
     }
   }, function (err) {
     function finalize (err) {
+      if (balanceBatchStarted) {
+        try {
+          library.clientWs.endBalanceBatch(!err && Boolean(saveBlock));
```
al-onyxprotocol, line 504: **Edge case: unrelated pending changes are discarded with a failed batch.** Balance changes from unconfirmed-transaction merges that arrive *while* a block apply is in flight get queued into this block's batch (the batch depth is global on the ClientWs instance). If the block fails, or is a replay with `saveBlock=false`, those unrelated events are dropped along with the block's own changes. Consistent with the documented best-effort semantics, but during a long resync every balance event landing inside an apply window is swallowed.

**Recommendation:** acceptable as-is given the non-durability contract; optionally flush (rather than discard) entries whose address was not touched by the failed block, if this ever matters in practice.

modules/clientWs/transactionSubscription.js
```diff
@@ -134,6 +147,50 @@ class TransactionSubscription {
     return subscribed;
   }
 
+  /**
+   * Subscribes to supported public account balance fields.
+   * @param {...string} fields - Public account fields to subscribe to
+   * @return {boolean} Whether at least one supported field was subscribed
+   */
+  subscribeToBalances (...fields) {
```
al-onyxprotocol, line 155: **Hardening: no per-socket subscription cap.** A single socket can subscribe to an unbounded number of addresses (pre-existing), and with balance fields subscribed each such address now costs up to one extra account DB read per changed address per block. The per-address index bounds this well, but the public endpoint would benefit from a ceiling.

**Recommendation:** consider a configurable per-socket limit on subscribed addresses/fields (e.g. a few hundred), returning `false` beyond it — cheap to add here in `TransactionSubscription`.

al-onyxprotocol (commented):
Reviewed the full diff, ran the seven targeted suites locally (226 passing), ESLint, and `git diff --check` — all clean. The subscription indexes, balance-read batching, and error isolation are solid; I traced the `registerSubscription`/`removeSubscription` invariants and found no reachable stale-index leak. No blocking issues — below are minor remarks with recommendations.

modules/rounds.js
```diff
@@ -318,6 +324,10 @@ Rounds.prototype.tick = function (block, done) {
     // Stop round ticking
     __private.ticking = false;
 
+    if (!err) {
+      __private.emitRoundBalanceChanges(scope);
+    }
+
     if (scope.finishSnapshot) {
```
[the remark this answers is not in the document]
massivedev0: Fixed in 40a3ad89. Forward round processing now skips reward-balance scheduling when snapshot truncation finishes, and a regression test verifies that Snapshot finished produces no client balance publication.

modules/clientWs.js
massivedev0: Fixed in 40a3ad89. Any nested endBalanceBatch(false) now latches discard state until the outermost batch closes; the latch is reset afterward. Added regression coverage for the nested case.

modules/clientWs.js
massivedev0: Fixed in 40a3ad89. The pending entry is now initialized with only its fields set, and the newest getAccount accessor is assigned once afterward.

modules/clientWs.js
massivedev0: Documented in Adamant-im/docs@dae6da2. The best-effort section now explicitly notes that a rolled-back block re-applied within the duplicate-suppression window can be omitted.

modules/blocks/chain.js
massivedev0: No code change. Production block apply/rollback and public unconfirmed transaction processing are serialized through the same balancesSequence, so an unrelated account merge cannot interleave with an active block batch today. Keeping whole-batch discard also preserves atomic best-effort behavior if the batch fails.

modules/clientWs/transactionSubscription.js
massivedev0: No code change in this PR. A silent hard cap would change the pre-existing subscription contract, and this API has no acknowledgement for clients to detect partial rejection. A ceiling should be introduced separately as a configurable, documented resource policy with explicit client feedback; the current address index continues to prevent unrelated socket scans and account reads.

massivedev0:
Addressed the second review round in `40a3ad89`.

- suppress round reward balance scheduling when snapshot truncation finishes
- latch nested batch suppression until the outer balance batch closes
- remove the redundant pending account accessor assignment
- add regression coverage for snapshot suppression and nested batch discard
- document the block re-apply duplicate-suppression edge case in `Adamant-im/docs@dae6da2`

The per-socket subscription ceiling and the alleged unrelated-merge interleaving were left unchanged for the reasons documented in their threads: a new silent cap needs a separate compatible API/config contract, while current balance mutation paths share `balancesSequence` and cannot interleave with block application.

Validation:

- targeted ClientWs/rounds tests: 17 passing
- `npm run test:unit:fast`: 940 passing
- `npm run eslint`: passed
- `git diff --check`: passed

PR #266, opened 2026-07-16
Feat: Expose consensus schedules and delegate forged amounts

### LINKED ISSUE #234: [Enhancement] Expose consensus and milestone schedules in status APIs
## Summary

Extend the public node status APIs with machine-readable information about the currently active consensus upgrade, configured consensus activation heights, and the block reward milestone schedule.

The API must expose the node's effective runtime values after configuration defaults and startup overrides are merged. It must not expose the full node configuration or any sensitive values.

## Motivation

Clients, explorers, monitoring systems, and node operators currently cannot determine through the public API:

- which consensus upgrade is active at the node's current blockchain height;
- which activation heights the node is configured to use;
- the complete block reward milestone schedule.

For example, `GET /api/blocks/getMilestone` only returns the current milestone index. Consumers must otherwise duplicate constants or inspect node configuration and source code. This makes it harder to verify that nodes use network-compatible activation settings and to present future reward changes correctly.

### 1. Extend `GET /api/blocks/getStatus`

Add the latest activated consensus codename:

```json
{
  "success": true,
  "consensus": "fairSystem"
}
```

Semantics:

- return `null` before the first configured consensus activation;
- return the codename with the greatest activation height that is less than or equal to the current blockchain height;
- activation at height `H` is active for block height `H`;
- use the node's effective runtime activation configuration;
- preserve every existing response field.

Consensus upgrades are cumulative. This field identifies the latest activated upgrade and does not imply that earlier upgrades became inactive.

### 2. Extend `GET /api/node/status`

Add the same latest activated consensus codename to `network`:

```json
{
  "network": {
    "height": 12345678,
    "consensus": "fairSystem"
  }
}
```

Add a top-level consensus schedule:

```json
{
  "consensusSchedule": {
    "activationHeights": {
      "fairSystem": 809,
      "spaceship": 1819
    }
  }
}
```

The numbers above are illustrative testnet values. Every node must return its own effective values after applying `config.default.json`, the selected config file, `--config-overrides`, and `--config-set`.

Add a top-level reward milestone schedule. Suggested response shape:

```json
{
  "milestoneSchedule": {
    "offset": 2000000,
    "distance": 6300000,
    "milestones": [
      50000000,
      45000000,
      40000000,
      35000000,
      30000000,
      25000000,
      20000000,
      15000000,
      10000000
    ]
  }
}
```

Reward values remain expressed in the node's current base unit (`1 ADM = 100000000`).

### 3. Documentation and schema

Update the companion artifacts together with the node implementation:

- Node API documentation: https://github.com/Adamant-im/docs
- API schema: https://github.com/Adamant-im/adamant-schema

Document field types, units, activation boundary semantics, network-specific values, and backward compatibility.

- Derive the active consensus codename from the same effective activation-height map used by `logic/consensus/consensus.js`.
- Avoid duplicating activation logic independently in the two API modules.
- Expose only an allowlisted public subset of configuration.
- Read the milestone schedule from `helpers/constants.js` rather than duplicating reward values in API code.
- Keep the response deterministic and based only on the applied blockchain height and effective startup configuration.

No consensus validation, delegate ordering, reward calculation, block serialization, or replay behavior should change.

## Acceptance criteria

- [ ] `GET /api/blocks/getStatus` returns `consensus`.
- [ ] `GET /api/node/status` returns `network.consensus`.
- [ ] Both endpoints return the same consensus codename for the same applied height.
- [ ] The pre-activation value is `null`.
- [ ] Activation is reported starting exactly at its configured activation height.
- [ ] `GET /api/node/status` returns `consensusSchedule.activationHeights`.
- [ ] Returned activation heights reflect all effective runtime overrides.
- [ ] `GET /api/node/status` returns `milestoneSchedule`.
- [ ] Existing response fields remain backward-compatible.
- [ ] No secrets, forging passphrases, database credentials, Redis credentials, peer access controls, or unrelated config values are exposed.
- [ ] Unit/API tests cover pre-activation, exact-boundary, post-activation, overridden activation heights, and milestone schedule output.
- [ ] ADAMANT node documentation is updated.
- [ ] `adamant-schema` is updated.

## Alternatives

Adding a generic endpoint that returns the complete node configuration was considered and rejected because the configuration contains sensitive operational data. Explicit allowlisted status fields are safer and provide a stable public API contract.

## Pull request

- #266

### LINKED ISSUE #265: [Enhancement] Include lifetime forged amount in delegate responses
## Summary

Include each delegate's lifetime forged amount in `GET /api/delegates` and `GET /api/delegates/get` responses as a decimal string field named `forged`.

## Motivation

Consumers that display the active delegate list currently need one `GET /api/delegates/forging/getForgedByAccount` request per delegate. ADAMANT Explorer therefore makes up to 101 additional requests to populate its `Forged` column, even though the Node already keeps lifetime delegate fees and rewards in `mem_accounts`.

This is needed by [Adamant-im/adamant-explorer#14](https://github.com/Adamant-im/adamant-explorer/issues/14) and should be available in the Node API and typed SDK before the next `adamant-api` npm release.

## Detailed description

Return `forged` for every delegate object using the same lifetime semantics as `GET /api/delegates/forging/getForgedByAccount` without `start` or `end`:

```text
forged = fees + rewards
```

The value must be returned as a base-10 integer string in sats, matching the existing forged-statistics endpoint and avoiding unsafe JavaScript integer conversion.

The list and single-delegate endpoints should expose the same delegate shape. Existing sorting, pagination, active-delegate limits, caching, and response fields must remain compatible.

## Alternatives

- Keep 101 per-delegate forged-statistics requests in Explorer: functionally correct but unnecessarily increases request count and public Node load
- Add a separate bulk endpoint: avoids fan-out but duplicates delegate-list lookup and introduces another contract when the existing delegate DTO is the natural response shape
- Reconstruct lifetime forged totals from blocks: more expensive and can diverge from the Node's canonical `fees + rewards` accounting

`Delegates.prototype.getDelegates()` already performs one `modules.accounts.getAccounts()` read from `mem_accounts`. The same rows contain `fees` and `rewards`, and `getForgedByAccount` already calculates lifetime forged amount by adding those fields.

- Add `fees` and `rewards` to the existing account-field projection in `modules/delegates.js`
- Derive `delegate.forged` with `bignum` and return it as a string
- Avoid any per-delegate database query or block-table aggregation
- Add unit and API tests for both `/api/delegates` and `/api/delegates/get`
- Verify `forged` matches `/api/delegates/forging/getForgedByAccount` for the same delegate
- Update the ADAMANT API schema and `DelegateDto` example/type in `Adamant-im/adamant-schema`
- Regenerate/sync `adamant-api-jsclient` types and add SDK coverage before publishing the npm package

Expected Node overhead is small: the existing in-memory account query returns two additional `BIGINT` columns per delegate, followed by at most 101 additions and one extra string in each delegate response. This replaces up to 101 HTTP requests and account lookups per Explorer refresh.

## Pull request

- #266

### LINKED ISSUE #267: [Bug] getNextForgers returns previous-round schedule at round boundaries
## Summary

`GET /api/delegates/getNextForgers` can return the previous round's delegate schedule while the last block is a round-boundary block. During that interval, the first reported next forger may differ from the delegate that the forging path selects for the next block.

## Steps to reproduce

1. Run the Node until the last block height is divisible by 101, for example height 101
2. Call `GET /api/delegates/getNextForgers?limit=1` before block 102 is accepted
3. Compare the returned delegate with the delegate selected for forging block 102

The mismatch is also visible directly in `modules/delegates.js`:

- `getNextForgers` calls `generateDelegateList(currentBlock.height)`
- The forging path calls `getBlockSlotData(currentSlot, lastBlock.height + 1, ...)`, which calls `generateDelegateList(height)`
- `modules.rounds.calc(height)` uses `Math.ceil(height / 101)`

At height 101, the endpoint therefore generates the round 1 list while the forging path generates the round 2 list for block 102.

## Expected behavior

`getNextForgers` should use the same height and delegate-list semantics as the forging path for the next block. At a round boundary, the first returned delegate must match the delegate eligible to forge the next block.

Add a unit regression test with the last block at a height divisible by 101. The test should verify that the endpoint requests or derives the schedule for `currentBlock.height + 1` and returns the same next-slot delegate as the forging path.

The semantics of `limit > 1` near a round boundary should also be documented or covered by tests because a response generated from one static delegate list may span a subsequent height-based round transition.

## Actual behavior

The endpoint generates the delegate list from `currentBlock.height`. When the last block closes a round, this uses the previous round seed until the next block is accepted, even though the next forging attempt already uses `lastBlock.height + 1` and the new round seed.

This is an observation API mismatch; the forging and consensus paths themselves already use the next block height and should remain unchanged.

## Screenshots or videos

Not applicable

## Environment

- ADAMANT Node v0.10.2
- Reproduced from the current `dev` implementation in `modules/delegates.js`
- Endpoint: `GET /api/delegates/getNextForgers`

al-onyxprotocol:
## Description

Expose the node's effective consensus activation and block reward schedules through the public status APIs, include each delegate's lifetime forged amount in delegate list and single-delegate responses, and correct the next-forger schedule at round boundaries.

The status changes let clients, explorers, monitors, and operators discover:

- the latest cumulative consensus upgrade active at the applied block height;
- the effective activation heights after configuration defaults and runtime overrides are merged;
- the complete block reward milestone schedule.

The delegate changes:

- add `forged` as a base-10 integer string calculated from the existing lifetime `fees + rewards` account fields, without per-delegate database requests;
- make `GET /api/delegates/getNextForgers` use the delegate schedule for `currentBlock.height + 1`, matching the forging path at round boundaries;
- return the standard `Blockchain is loading` API error instead of evaluating a missing chain-tip height.

The new response fields are additive. The next-forger fix changes only the observation API's previously incorrect round-boundary output. Consensus validation, delegate ordering, reward calculation, block serialization, replay behavior, and storage are unchanged.

## Related issue

Closes #234
Closes #265
Closes #267

## External links (optional)

- Documentation companion PR: Adamant-im/docs#37
- OpenAPI companion PR: Adamant-im/adamant-schema#51

Not applicable; this PR changes JSON API responses only.

## Breaking changes

No contract-breaking fields are removed or renamed:

- `GET /api/blocks/getStatus` adds `consensusCodeName`;
- `GET /api/node/status` adds `network.consensusCodeName`, `consensusSchedule`, and `milestoneSchedule`;
- `GET /api/delegates` and `GET /api/delegates/get` add `forged`;
- `GET /api/delegates/getNextForgers` preserves its response shape while correcting the schedule used at round boundaries.

Clients supporting older node versions should continue to tolerate the absence of the additive fields.

## How to test

Run the focused unit tests:

```sh
PG_NATIVE=false npm run test:single -- test/unit/modules/blocks.api.js test/unit/modules/delegates.js
```

Run the broader fast unit suite:

```sh
PG_NATIVE=false npm run test:unit:fast
```

Observed results for the latest review follow-up:

- focused blocks/delegates unit tests: 49 passing;
- fast unit suite: 954 passing;
- ESLint for all touched files: passed;
- `git diff --check`: passed.

Earlier focused API coverage for the additive status and forged fields passed with 5 tests. The full long-running unit and API suites were not run because the focused and fast suites directly cover the changed paths.

`PG_NATIVE=false` was used because the local `libpq` addon targets Node ABI 127 while the active Node runtime expects ABI 147. The changes do not affect database driver behavior.

## Notes for reviewers

- `Consensus#getActiveCodeName()` is the single deterministic source used by both status endpoints. Activation begins exactly at height `H`, and equal-height entries use a deterministic codename tie-break.
- `consensusCodeName` avoids confusion with the existing numeric `loader.consensus` peer-agreement percentage and is synchronized with the companion docs and OpenAPI PRs.
- Status endpoints avoid invoking consensus and reward helpers until a finite applied block height is available.
- `getNextForgers` uses the next block height for one chain-tip snapshot; clients should refresh after another block is accepted.
- The consensus schedule is returned as a copy of the effective allowlisted activation-height map; unrelated configuration and secrets are not exposed.
- Delegate `forged` uses the same lifetime `fees + rewards` semantics and decimal-string representation as `getForgedByAccount`, without N+1 queries.
- Security, privacy, protocol validation, storage, serialization, replay, reward settlement, and delegate ranking behavior are unchanged.

## Checklist

- [x] Code is formatted
- [x] Documentation updated (if needed)
- [x] Changes tested in all relevant environments

### DIFF
--- a/modules/blocks/api.js
+++ b/modules/blocks/api.js
@@ -22,15 +22,17 @@ __private.blockReward = new BlockReward();
  * @param {Block} block
  * @param {ZSchema} schema
  * @param {Sequence} dbSequence
+ * @param {Consensus} consensus
  */
-function API (logger, db, block, schema, dbSequence) {
+function API (logger, db, block, schema, dbSequence, consensus) {
   library = {
     logger: logger,
     db: db,
     schema: schema,
     dbSequence: dbSequence,
     logic: {
-      block: block
+      block: block,
+      consensus: consensus
     }
   };
   self = this;
@@ -298,6 +300,7 @@ API.prototype.getStatus = function (req, cb) {
     broadhash: modules.system.getBroadhash(),
     epoch: constants.epochTime,
     height: lastBlock.height,
+    consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
     fee: library.logic.block.calculateFee(),
     milestone: __private.blockReward.calcMilestone(lastBlock.height),
     nethash: modules.system.getNethash(),
--- a/modules/blocks/api.js
+++ b/modules/blocks/api.js
@@ -295,6 +295,9 @@ API.prototype.getStatus = function (req, cb) {
   }
 
   var lastBlock = modules.blocks.lastBlock.get();
+  if (!Number.isFinite(lastBlock.height)) {
+    return setImmediate(cb, 'Blockchain is loading');
+  }
 
   return setImmediate(cb, null, {
     broadhash: modules.system.getBroadhash(),
--- a/modules/blocks/api.js
+++ b/modules/blocks/api.js
@@ -300,7 +300,7 @@ API.prototype.getStatus = function (req, cb) {
     broadhash: modules.system.getBroadhash(),
     epoch: constants.epochTime,
     height: lastBlock.height,
-    consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
+    consensusCodeName: library.logic.consensus.getActiveCodeName(lastBlock.height),
     fee: library.logic.block.calculateFee(),
     milestone: __private.blockReward.calcMilestone(lastBlock.height),
     nethash: modules.system.getNethash(),
--- a/modules/node.js
+++ b/modules/node.js
@@ -34,7 +34,10 @@ function Node (cb, scope) {
     bus: scope.bus,
     nonce: scope.nonce,
     build: scope.build,
-    logic: scope.logic,
+    logic: {
+      block: scope.logic.block,
+      consensus: scope.logic.consensus || scope.consensus
+    },
     lastCommit: scope.lastCommit,
     config: {
       peers: scope.config.peers,
@@ -42,7 +45,6 @@ function Node (cb, scope) {
       wsClient: scope.config.wsClient
     }
   };
-  library.logic.consensus = scope.logic.consensus || scope.consensus;
   self = this;
 
   setImmediate(cb, null, self);
@@ -107,6 +109,7 @@ Node.prototype.shared = {
    */
   getStatus: function (req, cb) {
     var lastBlock = modules.blocks.lastBlock.get();
+    var hasBlockHeight = Number.isFinite(lastBlock.height);
     var nodeTimestampMs = slots.getTimeMs();
     var wsClientOptions = {
       enabled: false
@@ -131,7 +134,7 @@ Node.prototype.shared = {
             broadhash: modules.system.getBroadhash(),
             epoch: constants.epochTime,
             height: lastBlock.height,
-            consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
+            consensusCodeName: hasBlockHeight ? library.logic.consensus.getActiveCodeName(lastBlock.height) : null,
             fee: library.logic.block.calculateFee(),
             milestone: lastBlock.height ? __private.blockReward.calcMilestone(lastBlock.height) : undefined,
             nethash: modules.system.getNethash(),
--- a/modules/node.js
+++ b/modules/node.js
@@ -42,6 +42,7 @@ function Node (cb, scope) {
       wsClient: scope.config.wsClient
     }
   };
+  library.logic.consensus = scope.logic.consensus || scope.consensus;
   self = this;
 
   setImmediate(cb, null, self);
@@ -130,11 +131,20 @@ Node.prototype.shared = {
             broadhash: modules.system.getBroadhash(),
             epoch: constants.epochTime,
             height: lastBlock.height,
+            consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
             fee: library.logic.block.calculateFee(),
             milestone: lastBlock.height ? __private.blockReward.calcMilestone(lastBlock.height) : undefined,
             nethash: modules.system.getNethash(),
             reward: lastBlock.height ? __private.blockReward.calcReward(lastBlock.height) : undefined,
             supply: lastBlock.height ? __private.blockReward.calcSupply(lastBlock.height) : undefined
+          },
+          consensusSchedule: {
+            activationHeights: library.logic.consensus.getActivationHeights()
+          },
+          milestoneSchedule: {
+            offset: constants.rewards.offset,
+            distance: constants.rewards.distance,
+            milestones: constants.rewards.milestones.slice()
           },
           version: {
             build: library.build,
--- a/test/api/node.js
+++ b/test/api/node.js
@@ -1,0 +1,39 @@
+'use strict';
+
+var node = require('./../node.js');
+
+describe('GET /api/node/status', function () {
+  it('should expose effective consensus and milestone schedules', function (done) {
+    node.get('/api/node/status', function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.nested.property('network.height').that.is.a('number');
+      node.expect(res.body).to.have.nested.property('network.consensus').that.satisfy(function (consensus) {
+        return consensus === null || typeof consensus === 'string';
+      });
+
+      node.expect(res.body).to.have.property('consensusSchedule').that.has.all.keys('activationHeights');
+      node.expect(res.body.consensusSchedule.activationHeights).to.deep.equal(node.config.consensusActivationHeights);
+
+      node.expect(res.body).to.have.property('milestoneSchedule');
+      node.expect(res.body.milestoneSchedule).to.deep.equal({
+        offset: node.constants.rewards.offset,
+        distance: node.constants.rewards.distance,
+        milestones: node.constants.rewards.milestones
+      });
+
+      done();
+    });
+  });
+
+  it('should report the same consensus as the blocks status endpoint', function (done) {
+    node.get('/api/node/status', function (err, nodeStatus) {
+      node.expect(nodeStatus.body).to.have.property('success').to.be.true;
+
+      node.get('/api/blocks/getStatus', function (err, blocksStatus) {
+        node.expect(blocksStatus.body).to.have.property('success').to.be.true;
+        node.expect(nodeStatus.body.network.consensus).to.equal(blocksStatus.body.consensus);
+        done();
+      });
+    });
+  });
+});
--- a/test/api/node.js
+++ b/test/api/node.js
@@ -7,8 +7,8 @@ describe('GET /api/node/status', function () {
     node.get('/api/node/status', function (err, res) {
       node.expect(res.body).to.have.property('success').to.be.true;
       node.expect(res.body).to.have.nested.property('network.height').that.is.a('number');
-      node.expect(res.body).to.have.nested.property('network.consensus').that.satisfy(function (consensus) {
-        return consensus === null || typeof consensus === 'string';
+      node.expect(res.body).to.have.nested.property('network.consensusCodeName').that.satisfy(function (consensusCodeName) {
+        return consensusCodeName === null || typeof consensusCodeName === 'string';
       });
 
       node.expect(res.body).to.have.property('consensusSchedule').that.has.all.keys('activationHeights');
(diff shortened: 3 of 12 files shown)


adamantmm (approved):

modules/node.js
```diff
@@ -42,6 +42,7 @@ function Node (cb, scope) {
       wsClient: scope.config.wsClient
     }
   };
+  library.logic.consensus = scope.logic.consensus || scope.consensus;
```
adamant-al: `library.logic` is assigned `scope.logic` by reference on line 37, so this assignment writes into the `scope.logic` object shared by every module. In production it is a no-op (consensus is already there), but in tests it silently injects the test's consensus instance into the shared `modulesLoader` scope, where it can leak into other suites initialized from the same scope object.

`modules/delegates.js:51` already models the cleaner pattern — resolving the `scope.logic.consensus || scope.consensus` fallback into a locally-owned object instead of mutating the shared one. Suggest mirroring that here.

modules/node.js
```diff
@@ -130,12 +131,21 @@ Node.prototype.shared = {
             broadhash: modules.system.getBroadhash(),
             epoch: constants.epochTime,
             height: lastBlock.height,
+            consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
```
adamant-al: `consensus` is the only unguarded field among its siblings: `milestone`, `reward`, and `supply` all use `lastBlock.height ? … : undefined`, acknowledging a reachable state where the height is falsy. If `lastBlock.height` is `undefined` in that window, `getActiveCodeName(undefined)` falls through to `getCurrentHeight()`, which can throw `'Consensus height source is unavailable'` synchronously inside `getStatus` — surfacing as an unstructured 500 instead of the graceful degradation the neighboring fields chose.

Suggest matching the local convention:

```js
consensus: lastBlock.height ? library.logic.consensus.getActiveCodeName(lastBlock.height) : null,
```

(The blocks endpoint is fine as-is — it is gated by `__private.loaded` and its existing fields are equally unguarded.)

modules/blocks/api.js
```diff
@@ -298,6 +300,7 @@ API.prototype.getStatus = function (req, cb) {
     broadhash: modules.system.getBroadhash(),
     epoch: constants.epochTime,
     height: lastBlock.height,
+    consensus: library.logic.consensus.getActiveCodeName(lastBlock.height),
```
adamant-al: Naming consideration for both endpoints: `/api/node/status` now carries two same-named fields with unrelated meanings — `loader.consensus` (peer sync-agreement percentage, a number) and `network.consensus` (protocol upgrade code name, a string). The same divergence appears here: this new top-level `consensus` is the code name, while `consensus` historically means the percentage in this codebase.

It is additive and the companion docs/schema PRs help, but something like `network.consensusCodeName` (and `consensusCodeName` here) would spare client authors a footgun. Worth deciding now — renaming after release is exactly the breaking change this PR carefully avoided.

test/api/node.js
```diff
@@ -0,0 +1,39 @@
...
+  it('should expose effective consensus and milestone schedules', function (done) {
+    node.get('/api/node/status', function (err, res) {
+      node.expect(res.body).to.have.property('success').to.be.true;
+      node.expect(res.body).to.have.nested.property('network.height').that.is.a('number');
+      node.expect(res.body).to.have.nested.property('network.consensus').that.satisfy(function (consensus) {
+        return consensus === null || typeof consensus === 'string';
+      });
+
+      node.expect(res.body).to.have.property('consensusSchedule').that.has.all.keys('activationHeights');
+      node.expect(res.body.consensusSchedule.activationHeights).to.deep.equal(node.config.consensusActivationHeights);
```
adamant-al, line 15: This compares the endpoint's *merged* schedule (config defaults + overrides) against the raw `node.config.consensusActivationHeights`. It passes because `test/config.json` currently enumerates every key in `config.default.json`, but adding a new upgrade to defaults without touching the test config will fail this test with a confusing diff.

Arguably that is a useful tripwire that forces the test config to stay in sync — just flagging it so it is intentional rather than accidental.

adamant-al (commented):
Reviewed the changes with full source context. Clean, additive API change overall: `getActivationHeights()` returns a defensive copy, `forged` rides the existing in-memory delegates query (no N+1) and provably matches `getForgedByAccount` semantics (same account fields, same bignum string math; pg returns `int8` as strings and there are no `setTypeParser` overrides in the repo). Verified the claimed unit tests locally: 23 + 4 + 34 passing. Production wiring is safe — consensus is created in app.js's `logic` section and the HTTP server only listens after `bindModules` runs, so there is no unbound-consensus window.

No critical issues. Four low-severity suggestions inline; #1 and #2 are worth a small follow-up commit, and the naming one deserves an explicit decision before merge since it is the one thing that is hard to change later.

modules/node.js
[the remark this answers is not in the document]
al-onyxprotocol: Fixed in `0ffccf06`. `/api/node/status` now checks `Number.isFinite(lastBlock.height)` and returns `network.consensusCodeName: null` until an applied height is available, so the field cannot fall back to an unrelated height source or throw during early startup.

modules/blocks/api.js
al-onyxprotocol: Implemented in `0ffccf06`, with companion updates in Adamant-im/docs@`0e48551` and Adamant-im/adamant-schema@`52295b9`. The new string field is now `consensusCodeName` / `network.consensusCodeName`, while the existing numeric `loader.consensus` peer-agreement percentage remains unchanged.

